PR #117, opened 2015-01-06
caching tutorial

jardakotesovec:
Just finished first version of caching tutorial. Any feedback (including english corrections) appreciated.

Thanks!

### DIFF
--- a/lib/tutorials/caching.md
+++ b/lib/tutorials/caching.md
@@ -1,13 +1,14 @@
 ## Client side caching
 
-Http protocol provides several mechanisms how to cache resources in browser via http headers. To decide which headers are suitable for your use case check Google developers [guide](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching) or other [resources](https://www.google.com/search?q=HTTP+caching). This tutorial provides overview how to use these techniques in hapi.
+HTTP protocol provides several different HTTP headers to control how browsers cache resources. To decide which headers are suitable for your use case check Google developers [guide](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching) or other [resources](https://www.google.com/search?q=HTTP+caching). This tutorial provides overview how to use these techniques in hapi.
 
 ### Cache-Control
 
-`Cache-Control` header tells browser and any intermediate cache if resource is cacheable and for how long. For example `Cache-Control:max-age=30, must-revalidate, private` means that browser can cache resource for 30 seconds and `private` means that it should not be cached by intermediate caches, only by browser. `must-revalidate` means that once it expires it has to request the resource again.
+The `Cache-Control` header tells the browser and any intermediate cache if resource is cacheable and for how long. For example `Cache-Control:max-age=30, must-revalidate, private` means that the browser can cache resource for 30 seconds and `private` means that it should not be cached by intermediate caches, only by the browser. `must-revalidate` means that once it expires it has to request the resource again.
 
 Lets see how we can set this header in hapi:
 
+**Listing 1** Setting Cache-Control header
 ```javascript
 server.route({
    path: '/hapi/{ttl?}',
@@ -33,14 +34,15 @@ Example also illustrates that `expiresIn` value can be overridden with `ttl(msec
 See [route-options](http://hapijs.com/api#route-options) for more options.
 ### Last-Modified
 
-In some cases server can provide information when resources was last modified. Good example is [inert](https://github.com/hapijs/inert) module, which is responsible for serving static files in hapi (`file` and `directory` handler). Last modification timestamp is read for every requested file and send as `Last-Modified` header.
+In some cases the server can provide information when resources was last modified. When using either the `file` or `directory` route configuration option, the underlying module, [inert](https://github.com/hapijs/inert) add `Last-Modified` for every requested file.
 
-When `Last-Modified` is set on response hapi compares it with `If-Modified-Since` coming from client to decide if should return `304`.
+When `Last-Modified` is set on response hapi compares it with `If-Modified-Since` coming from client to decide if the response status code should be `304`.
 
 
 Assuming `lastModified` is Date object you can set this header via [response object](http://hapijs.com/api#response-object) interface.
 
-```
+**Listing 2** Setting Last-Modified header
+```javascript
    reply(result).header('Last-Modified', lastModified.toUTCString());
 ```
 
@@ -48,11 +50,12 @@ There is one more example using `Last-Modified` in [last section](#client-and-se
 
 ### ETag
 
-ETag is alternative to `last-modified` where server provides token (usually checksum of resource) instead of last modified timestamp. Browser in next request sends this ETag in `If-None-Match` header. Server compares it with new `ETag` and return `304` if it's the same.
+The ETag header is alternative to `last-modified` where server provides token (usually checksum of resource) instead of last modified timestamp. Browser in next request sends this ETag in `If-None-Match` header. Server compares it with new `ETag` and return `304` if it's the same.
 
 You only need to set `ETag` in your handler via `etag(tag, options)` function:
 
-```
+**Listing 3** Setting ETag header
+```javascript
    reply(result).etag('xxxxxxxxx');
 ```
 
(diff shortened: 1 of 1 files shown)


nlf:
Wow! At first glance, this is awesome! Great work!

When I get some time I'll review it more closely and we'll get it merged. Thank you for this!

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
+## Client side caching
+
+Http protocol provides several mechanisms how to cache resources in browser via http headers. To decide which headers are suitable for your use case check Google developers [guide](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching) or other [resources](https://www.google.com/search?q=HTTP+caching). This tutorial provides overview how to use these techniques in hapi.
```
arb: > mechanisms how to cache resources in browser via http headers
maybe change to "different Http headers to control how browsers cache resources."

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Cache-Control
+
+`Cache-Control` header tells browser and any intermediate cache if resource is cacheable and for how long. For example `Cache-Control:max-age=30, must-revalidate, private` means that browser can cache resource for 30 seconds and `private` means that it should not be cached by intermediate caches, only by browser. `must-revalidate` means that once it expires it has to request the resource again.
```
arb: "The `Cache-Contol` header"

I think you need to put "the" before the word "browser" in this whole paragraph.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+      }
+   },
+   config: {
+      cache: {
+         expiresIn: 30 * 1000,
+         privacy: 'private'
+      }
+   }
+});
+```
```
arb: I generally name my code examples to I can refer to them easier; IE **Listing 1**.

lib/tutorials/caching.md
nlf: In addition to that suggestion, HTTP should always be in all caps

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+Example also illustrates that `expiresIn` value can be overridden with `ttl(msec)` function using [response object](http://hapijs.com/api#response-object) interface.
+
+See [route-options](http://hapijs.com/api#route-options) for more options.
+### Last-Modified
+
+In some cases server can provide information when resources was last modified. Good example is [inert](https://github.com/hapijs/inert) module, which is responsible for serving static files in hapi (`file` and `directory` handler). Last modification timestamp is read for every requested file and send as `Last-Modified` header.
```
arb: - "The server"
- "A good example"

Maybe reward the sentence about inert:
When using either the `file` or `directory` route configuration option, the underlying module, inert.... bla bla bla.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+When `Last-Modified` is set on response hapi compares it with `If-Modified-Since` coming from client to decide if should return `304`.
```
arb: "if the response status code should be 304."

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+Assuming `lastModified` is Date object you can set this header via [response object](http://hapijs.com/api#response-object) interface.
```
arb: Is this something the average person will need to know? Seems like something we could cut.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+```
+   reply(result).header('Last-Modified', lastModified.toUTCString());
+```
+
+There is one more example using `Last-Modified` in [last section](#client-and-server-caching).
+
+### ETag
+
+ETag is alternative to `last-modified` where server provides token (usually checksum of resource) instead of last modified timestamp. Browser in next request sends this ETag in `If-None-Match` header. Server compares it with new `ETag` and return `304` if it's the same.
```
arb: The "ETag" header is...

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+```
+   reply(result).etag('xxxxxxxxx');
+```
+
+Check `options` parameter for `etag` function on [api](http://hapijs.com/api#response-object) page for more details.
+
+
+## Server side caching
+
+hapi provides powerful server side caching via [catbox](https://www.github.com/hapijs/catbox) and makes using cache very convenient. This tutorial help you to understand how catbox is utilized inside hapi and how you can use it. Catbox also can be used out of the hapi, but that is not covered here.
```
arb: maybe how you can "leverage" it. Sounds fancier :grinning:

arb: Maybe remove the sentence about using catbox outside of hapi. Doesn't really seem helpful here.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+Catbox has two interfaces.
+
+
+### Client
+[Client](https://github.com/hapijs/catbox#client) is low-level interface, which basically allows you set/get key-value pairs. Its initialized with one of the available adapters ([Memory](https://github.com/hapijs/catbox-memory), [Redis](https://github.com/hapijs/catbox-redis), [mongoDB](https://github.com/hapijs/catbox-mongodb), [Memcached](https://github.com/hapijs/catbox-memcached), [Riak](https://github.com/DanielBarnes/catbox-riak), [Amazon S3](https://github.com/fhemberger/catbox-s3)).
```
arb: Should we include Amazon S3? It's technically not an "official" plugin.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+       }
+    ]
+});
+
+server.connection({
+    port: 8000
+});
+```
+
+In this example two [clients](https://github.com/hapijs/catbox#client) were defined: `'mongoCache'`, `'redisCache'` and also default one without name with [memory](https://github.com/hapijs/catbox-memory) adapter. You can replace default client by omitting `name` property. `partition` property tells adapter how cache should be named (`'catbox'` by default). In case of [mongoDB](http://www.mongodb.org/) it is database name and in case of [redis](http://redis.io/) it is used as key prefix.
```
arb: Maybe reword to something like:
In Listing 2, we've defined two catbox clients; mongoCache and redisCache. Including the default memory cache created by hapi, there are three available cache clients.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+    }
+});
+
+server.route({
+    path: '/add/{a}/{b}',
+    method: 'GET',
+    handler: function (request, reply) {
+
+        var id = request.params.a + ':' + request.params.b;
+        sumCache.get({id: id, a: request.params.a, b: request.params.b}, function (err, result) {
```
arb: Don't you only need the `id` key here in this object? I don't think `a` and `b` do anything?

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+server.connection({
+    port: 8000
+});
+```
+
+In this example two [clients](https://github.com/hapijs/catbox#client) were defined: `'mongoCache'`, `'redisCache'` and also default one without name with [memory](https://github.com/hapijs/catbox-memory) adapter. You can replace default client by omitting `name` property. `partition` property tells adapter how cache should be named (`'catbox'` by default). In case of [mongoDB](http://www.mongodb.org/) it is database name and in case of [redis](http://redis.io/) it is used as key prefix.
+
+### Policy
+[Policy](https://github.com/hapijs/catbox#policy) is high-level interface that you will be using. Lets pretend we are dealing with something more complicated than adding two numbers and we want to cache the results. [server.cache](http://hapijs.com/api#servercacheoptions) method creates new [policy](https://github.com/hapijs/catbox#policy), which is than used in handler. `id` object has mandatory key `id`, which is a unique item identifier string.
```
arb: Maybe remove the last sentence and create a new paragraph explaining `sumCache.get` below the code. 

It might be worth explaining what the `generateFunc` does (like why the signature is the way it is) as well.

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+In addition to **partitions** there are **segments** that allows to isolate caches within one [client](https://github.com/hapijs/catbox#client). If you want to cache results from two different methods, you usually don't want mix up the results. In [mongoDB](http://www.mongodb.org/) `segment` represents collection and in [redis](http://redis.io/) it's additional prefix to `partition` prefix. Default value for `segment` is provided when [server.cache](http://hapijs.com/api#servercacheoptions) is called inside plugin - `'!pluginName'` or used via [server methods](http://hapijs.com/tutorials/server-methods) - `'#methodName'`. If you have use case for multiple policies sharing one segment there is [shared](http://hapijs.com/api#servercacheoptions) option.
+
+### Server methods
+But it can get better than that! In 95% cases you will use [server methods](http://hapijs.com/tutorials/server-methods) for caching purposes, because it reduces boilerplate to minimum. Lets do the same with server methods:
+
+```javascript
+server.method('sum', add, { cache: { cache: 'mongoCache', expiresIn: 30 * 1000 } });
```
arb: For completeness, maybe include the `add` function here as well?

lib/tutorials/caching.md
```diff
@@ -0,0 +1,187 @@
...
+
+In this case we use `cached.stored` timestamp to set `last-modified` header.
+
+```javascript
+server.route({
+    path: '/add/{a}/{b}',
+    method: 'GET',
+    handler: function (request, reply) {
+
+        server.methods.sum(request.params.a, request.params.b, function (err, result, cached, report) {
```
arb: You could probably add these two parameters to the previous example so you didn't need to create a whole new one.

lib/tutorials/caching.md
jardakotesovec: Do you mean whole Last-Modified section? There is good example how to use Last-Modified with server side caching that can be pretty useful (in last section).

lib/tutorials/caching.md
jardakotesovec: You have to pass the params somehow (thats how server-methods do it) to get function so `generateFunc` can access them and can to do something with it.

arb:
@arb to make second English pass.

PR #127, opened 2015-01-19
fixes code sample

jzelenkov:


nlf:
This was actually correct as it was, it's wise to not start your server until the callback from `server.register` has completed to be certain that your plugins have completed initializing.

PR #137, opened 2015-01-22
added view helpers tutorial

manonthemat:


nlf:
The line wrapping for the quotes is going to break copy+paste ability of the example, so that should probably get fixed.

Also, can we incorporate your example server and templates into the tutorial directly? I would prefer that it didn't rely on external resources if at all possible.

manonthemat:
Thanks for the feedback. How is this now?

nlf:
:+1: thanks!

PR #153, opened 2015-03-13
updated opentable logo with rebranded one

ryantomlinson:


nlf:
Thanks!

Also, I like the new logo. Very sleek.

PR #164, opened 2015-04-08
Update good-console example

gergoerdosi:
Good 6.0.0 introduced some breaking changes that should be reflected in the tutorial. Note: good-console has a new version (5.0.0) which is compatible with good 6.0.0, however it is not published yet. I suggest to publish it before merging this PR.

### DIFF
--- a/lib/tutorials/getting-started.md
+++ b/lib/tutorials/getting-started.md
@@ -106,7 +106,10 @@ server.register({
     options: {
         reporters: [{
             reporter: require('good-console'),
-            args:[{ log: '*', response: '*' }]
+            events: {
+                request: '*',
+                log: '*'
+            }
         }]
     }
 }, function (err) {


nlf:
good-console@5.0.0 was published, so merging this now. thanks!

lib/tutorials/getting-started.md
```diff
@@ -106,7 +106,10 @@ server.register({
     options: {
         reporters: [{
             reporter: require('good-console'),
-            args:[{ log: '*', response: '*' }]
+            events: {
+                request: '*',
```
AdriVanHoudt: Why did you remove response and added request to log events? see https://github.com/hapijs/hapijs.com/issues/168

gergoerdosi: It was changed accidentally. Copy-pasted the code from the release notes (https://github.com/hapijs/good/issues/323) to test it locally and then probably forgot to update it when sent the PR. Created a new one: https://github.com/hapijs/hapijs.com/pull/169

PR #170, opened 2015-04-15
Update Concrete logo

owenDods:
We noticed that the Concrete logo had an off-white background that didn't look great so we figured we'd send you a better one :)


owenDods:
Whoa, handled this a bit badly, but the net result should be ok

nlf:
Thanks!

Issue #166, opened 2015-04-15
Add a link to hapi.js in Action book


mtharrison:
Would you consider adding a link to the book that I'm writing? http://manning.com/harrison/

There's also an affiliate program that can be worked out with Manning, whereby the project can gain commissions.

hueniverse:
There are a few ways we can do this. One is to add a books section and list it there along with others in the future. Another is to use a chapter from the book in the tutorials and from that page link to the book where people can learn more. I am sure there are more ideas.

As for the commissions, I don't see a way for us to collect these since that will require someone to receive the funds and deal with taxes, as well as deal with funds in the first place. If Manning wants to donate these funds, they can handle it and then give the money to some related community cause that is setup to receive donations.

AdriVanHoudt:
I would do both. A general page with links to educational materials is helpful. And if the book writer/owner wants to do a PR to the tutorials with a good tutorial and link to the book for further info, I see that as a win win. 
As for the commissions, I would prefer it to go to a good cause (dev/js related) or Hapi Swag ;)

mtharrison:
I'd be very happy to contribute some new tutorial content. For the meanwhile, if we could also have a Books or possibly a more general external learning resources page, as suggested by @AdriVanHoudt, that would be great too. Would you like me to work on a PR for this?

I'll speak to Manning regarding the commissions, it'd be nice to see them go to a good cause.

AdriVanHoudt:
A PR never hurts ;)

Marsup:
Asking beforehand neither ;)

AdriVanHoudt:
true :+1:

nlf:
Personally, I love the idea of an external resources section. It's something that's been brought up in the past by people looking for somewhere to link to gists and the like. I also really like the idea of having a tutorial that links to the book. Let me know what I can do to facilitate your PR :+1:

nlf:
Note, see #119 for additional thoughts

mtharrison:
Thanks @nlf. I'll put something together this weekend.

nlf:
this is merged and the link is added, thanks again @mtharrison

mtharrison:
Brilliant, thank you for that!

Issue #119, opened 2015-01-07
List of useful gists


jardakotesovec:
Recently I came across few very useful gists, like using bell with auth-cookie together, using socket.io with auth-cookie. How about make new section on website where these examples could live similarly as plugins?

Gists have also star rating, that could be displayed via https://github.com/mdo/github-buttons .

Thoughts?

nlf:
I had the thought a while back of changing the 'Tutorials' nav link to 'Learn' and also having code snippets available, which I think is almost exactly what you're talking about here.

devinivy:
This would be great.  A huge benefit I see is that the code examples could demonstrate use of community plugins, whereas the tutorials sections is generally just for vanilla hapi code.

AdriVanHoudt:
:+1:

nlf:
resources section has been added

PR #172, opened 2015-04-18
Added Learn page with some starting content

mtharrison:
Please see issue #166 for some more context.

I've added a Learn page with some subsections, styling/navigation is based on the plugins page. In issue #119 @nlf said he'd consider folding tutorials into this section too. Not sure what the thoughts are on that now. Personally I think it's nice to have the tutorials a single click away from the homepage but can see the benefit of combining them too.

Please feel free to throw suggestions or more content for this section at me and I'll add some more commits.


AdriVanHoudt:
I'm not a fan of the title `learn` since that hints more to tutorials, I would rather go with `external resources` or something 
Also might want to add https://gist.github.com/hueniverse/a06f6315ea736ed1b46d to the gists

mtharrison:
An issue with that name is that it wouldn't fit well into the top navigation. I think it would be a shame to demote the new section to the footer nav only.

So the options I can see are:

1. Go with **External Resources** and demote to footer nav only
2. Go with **Learn** and merge in tutorials
3. Add the content from this new section under another title, **Need Help?** seems to make sense because it already links to Postmile as an example application and Make Me Hapi.

AdriVanHoudt:
Hmm how about just finding a new name? Like `Books` or `Resources`?

mtharrison:
A whole section for **Books** wouldn't work at the moment, with only 1 book. **Resources** might be better, but I still think **Learn** is more descriptive of the content being featured. **Resources** sounds like it could be anything really. But it's your call ultimately, I appreciate you letting me add my book in any way :)

AdriVanHoudt:
Good points. And this is in no way my call I'm just giving my opinion.

nlf:
maybe `Resources` is the better name for this, then it becomes easier to differentiate between internal and external tutorials too.. i think if you rename learn to resources, this is ready to go. thanks!

templates/resources.jade
```diff
@@ -0,0 +1,30 @@
...
+                p= item.description
+
+    .column.nav-secondary
+      nav(role='toc')
+        h3 Resources
+
+        ul.unstyled(role='categories')
+          each category, categoryName in resources.categories
+            li(role='category')
+              a(href='##{categoryName}')= categoryName
```
AdriVanHoudt: newline here (if I can be picky ^^)

mtharrison: Sure

nlf:
thanks!

PR #191, opened 2015-06-05
Add resource article on authentication and authorization.

poeticninja:


lib/resources.js
```diff
@@ -63,6 +63,10 @@ exports.categories = {
         'Using hapi.js with Socket.io': {
             url: 'http://matt-harrison.com/using-hapi-js-with-socket-io/',
             description: 'Tutorial on how to integrate Socket.io with hapi.js'
+        },
+        'Authentication and Authorization with Hapi': {
```
hueniverse: hapi is always lowercase.

PR #193, opened 2015-06-10
resized the clarify logo

caseysoftware:
It was tiny and looked out of place


nlf:
this one may be a little too big, i'll adjust it a bit and make it look right

PR #195, opened 2015-06-29
Add hapi-next under Utility

Pranay92:
This module allows user to add modularity to the route handlers by allowing long function to be divided into small independent functions and be called in any sequence. This is also integrated with Travis-CI and has tests to ensure stable releases.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -220,6 +220,10 @@ exports.categories = {
             url: 'https://github.com/poeticninja/hapi-named-routes',
             description: 'Add named routes to your view templates.'
         },
+        'hapi-next' : {
+            url: 'https://github.com/Pranay92/hapi-next',
+            description: 'Add modularity to your route handlers.'
+        },
         'hapi-node-postgres': {
             url: 'https://github.com/jedireza/hapi-node-postgres',
             description: 'Wrap requests with a pg connection. Uses connection pooling via `node-postgres`.'


hueniverse:
How is this different in functionality from prerequisites?

Pranay92:
I'm sorry I didn't understood your statement. Is that achievable by default in hapi without using any plugins? I was trying to ind something like that but couldn't get it and ended up implementing the plugin itself. Please let me know if that's what you meant?

hueniverse:
Take a look at the prerequisite feature in hapi and see if it does what you want.

nlf:
@Pranay92 have you had a chance to look at prerequisites? if you have and you feel like this plugin is still useful it would be cool if you could add a note to your readme saying what's different about it.

Pranay92:
Hi @nlf apologies for the delayed response. I've been using hapi-next in our project in quite a while now, and for me its more of an opinion to add everything at one place(eg. in your handler) and then control the flow(the way its done in hapi-next) while also having modularity. I've added a few more options in the latest release of hapi-next (like passing on error status when any of the series function fails), please have a look at the updated readme too.

Also, after reading the read me, if you aren't satisfied enough feel free to reject this PR or (preferably) accept this :)

Thank you for the patience.

nlf:
updated readme looks nice and good usage info, thanks!

PR #199, opened 2015-07-15
add hapi-server-session to plugins list

btmorex:
Versus existing plugins:

* Never attempts to store session values in a cookie, only the uuid v4 session id
* Uses a diffing strategy to determine whether the session is dirty or not, which makes a nicer interface

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -124,6 +124,10 @@ exports.categories = {
         }
     },
     'Session': {
+        'hapi-server-session': {
+            url: 'https://github.com/btmorex/hapi-server-session',
+            description: 'Simple server-side session support for hapi'
+        },
         yar: {
             url: 'https://github.com/hapijs/yar',
             description: 'A hapi session plugin and cookie jar'


hueniverse:
You might want to read http://hueniverse.com/2015/07/08/on-securing-web-session-ids/ to understand why your implementation isn't sufficiently secure. You can just add the signature option to the server state definition as a minimum.

btmorex:
Thanks for your reply.

After reading that post, both github issues, and the OWASP guidelines, I concluded:

1. That using an unsigned random session id is not "insecure".
2. That signing a random session id gives some useful features. But see below how these can be implemented without signing.

So, I've released another version that includes the ability to sign the cookie and add an expiration. I also added tests and documentation.

In response to a few of the points in the post:

* **Alarms:** Distinguishing expired vs. invalid session ids is a good argument for signing. However, we can achieve roughly the same thing by adding a creation date to the server-side session and keeping the cache expiration significantly longer than the cookie expiration. Until the cache expires, we can still distinguish the two cases.
* **Hygiene:** Expiration is another good argument for signing, but again, we can achieve the same by storing a creation date on the server-side session.
* **Kill Switch:** Rotating the signing key and flushing the session cache both act as a kill switch. For example, if we discover that the random number generation has been faulty, invalidating all session ids is easy regardless of whether we sign the cookie or not.

nlf:
this looks much better, thanks for taking such a thorough look at the security concerns. great plugin :+1:

Issue #198, opened 2015-07-14
Does a detailed example Chat app using Hapi.js, Socket.io & Redis qualify as a "Resource"?


nelsonic:
Hi, we made this example app: https://github.com/dwyl/hapi-socketio-redis-chat-example and are wondering if it qualifies for a inclusion in the resources of hapijs.com ...
http://hapijs.com/resources ??
Thanks!

nlf:
I would say it absolutely qualifies

nelsonic:
Thanks for confirming @nlf  (_hope you are well!)
@iteles can you please prepare the PR? :+1:

iteles:
Wonderful, will do! :blush:

sericaia:
:+1:

nlf:
closed via #201

iteles:
Thanks @nlf, we really appreciate it! :tada:

PR #201, opened 2015-07-19
Adds chat example app into resources section as per #198

iteles:
As discussed with @nlf and @nelsonic in issue #198, this PR adds a [real-time chat app example](https://github.com/dwyl/hapi-socketio-redis-chat-example) to the list of resources.

**Why is this a useful resource?**
+ Uses **hapi.js**! (most real-time node app examples out there use express)
+ End-to-end **tests**
+ Integrates socket.io
+ Uses redis **pub/sub** as an added bonus

### DIFF
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -60,6 +60,10 @@ exports.categories = {
         }
     },
     'Tutorials': {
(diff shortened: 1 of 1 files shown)


nlf:
perfect! thanks for this

Issue #144, opened 2015-02-10
milestone/changelog availability


csrl:
Currently it seems the best place to find changes to Hapi is in the milestone filter, eg.

https://github.com/hapijs/hapi/issues?q=milestone%3A8.1.0

It'd be nice to have the milestones available from http://hapijs.com/updates

nlf:
I like this, maybe it's a better idea to list milestones and their associated issues directly rather than rendering the CHANGELOG.md. This would allow us to not have to rely on the CHANGELOG being kept up to date, as well as making it super easy to link to the milestones from their headers.

Anyone feel like taking this on?

hueniverse:
I would LOVE to get rid of the changlog file and have the latest always sit here.

pon:
What is the current process for updating the CHANGELOG.md file?

hueniverse:
I have to manually run `npm changelog` in hapi and commit the change.

mtharrison:
Happy to take this on along with #171

hueniverse:
@mtharrison both assigned to you now.

nlf:
@mtharrison keep in mind #196 when working on this. would be nice if we had the ability to display changelogs for all the plugins as well, and not just hapi itself.

mtharrison:
I think this is resolved now with changelog updates so closing.

AdriVanHoudt:
only it does not have 9.0 atm because of the cache time

PR #207, opened 2015-07-25
Dynamic changelog generation using the Github API. Closes #171, #144

### LINKED ISSUE #171: Move changelog generation to site
Remove the changelog.md in hapi and generate it by the site every time there is a new release. Also, need to change the changes page layout to not center the changes to make it easier to read it.

mtharrison:
The implements the dynamically-generated changelog to replace the pulled in changelog.md from the hapi project.

It's loosely based off of the mdchangelog tool except rather than downloading all issues for all modules it downloads them on a per-milestone basis. Even though this means more HTTP requests overall initially, it allows finer-grained caching. Once we have all issues for a closed milestone, they shouldn't change.

This is a first attempt. I welcome any feedback. I've never used Jade before (tend to go with Handlebars) so I might have made a dog's dinner of that.

I've also altered the changelog layout. It is now left-aligned and plugins are indented . Here's how it looks:

screen shot 2015-07-25 at 12 34 25
## Notes

Each issue in the main project is checked against a regex (taken from mdchangelog): 
`/(?:update|upgrade) (\w+)\/(\w+) to (.*) from (.*)/i`to determine whether any related commits need to fetch from another repo's milestone.
This means any issue/PR that were updates that don't match this regex, won't have their dependency issues pulled in. There's a couple of ways to fix this:
- Just ensure all following updates follow the format "Upgrade hapijs/joi to 5.1.0 from 5.1.0"
- Allow several patterns using multiple regexes
- Make the regex much looser. Checking only for keywords (known hapi repo name) and things that look like version numbers

We could distinguish issues from PRs in the changelog now if we want somehow or include author.


AdriVanHoudt:
Some quick thoughts: 
* why not use https://github.com/hapijs/qs ?
* be more more consistent (use `callback` or `cb` everywhere instead of mixing them and better use of return, see https://gist.github.com/hueniverse/a06f6315ea736ed1b46d#return-callbacks)
* I would also do
 
```js
 var internals = {
    hapiOrg: '' ",
    ...
}
```
reads better imo
* Indent the issues of the dep updates maybe?
* Not sure if a requirement but Hapi internals use https://github.com/hapijs/items instead of Async

mtharrison:
I used `querystring` and `async` because they're already being used in the project. Same with `cb` as shadowed `callback` variable, for consistency with the existing code (see https://github.com/hapijs/hapijs.com/blob/master/lib/github.js#L272).

I can declare internals that way, if you think it's better, I'll add a commit.

AdriVanHoudt:
hmm @nlf (I assume you are the maintainer but it is not listed in the readme) can you elaborate on these questions I had?

And ok sure if it is consistent with the rest of the repo. 
Leaves:
* indent issues of dep updates? not sure it will look better
* would it be possible to highlight/link the dep in the issue title instead of in the `strong` below?

nlf:
i dunno, i kinda like this as it is. it reads well to me and seems fairly organized. the regex looks ok for now, and we can always tune it over time.

i'm gonna go ahead and merge this and we'll see what public response looks like.

AdriVanHoudt:
yeah I was a bit nitpicking, happens when the rest is good ;)

mtharrison:
Awesome. Does this close #144 too?

AdriVanHoudt:
can we put an id on the Changelog title so we can link to it directly? :P I want to tweet this :D

AdriVanHoudt:
Also I would center it

nlf:
yeah.. i'm kinda thinking maybe keep the column left aligned, but center the column itself. it does look a bit weird fully left justified like it is now

nlf:
also maybe let's tweak it so it only shows the last 5-10 releases, that page is super long

AdriVanHoudt:
hmm looks like it is centered there are just some super wide sentences

AdriVanHoudt:
maybe we can open an issue with everything that needs tweaking?

AdriVanHoudt:
How can I force rerender the page for testing?

nlf:
right now, restart the server and refresh. the data is cached in memory, but the template is not. so if you're just changing the template a hard refresh would be enough. for data you'll need a full server restart.

mtharrison:
I was using catbox-redis when I was working on it. And then just doing `FLUSHDB` to clear cache. Another reason I did that is that catbox-memory doesn't work with really long (like a year) times due to limitations with setTimeout.

lib/changelog.js
```diff
@@ -0,0 +1,237 @@
...
+var Querystring = require('querystring');
+
+var internals = {};
+
+internals.dependencyRegex = /(?:update|upgrade) (\w+)\/(\w+) to (.*) from (.*)/i;
+
+internals.hapiOrg = 'hapijs';
+
+internals.perPage = 100;
+
```
danielb2: two spaces between top level items. but why not just do

internals = {
  dependencyRegex: ...
  hapiOrg: 'hapijs',
  perPage: 100
};

lib/changelog.js
```diff
@@ -0,0 +1,237 @@
...
+                    }
+
+                    cb(null);
+                });
+            }
+
+            cb(null);
+        };
+
+        var getIssues = function (milestone, cb) {
```
danielb2: Instead of having this inside a giant function, you can put this into  `internals.checkForDependencies` and move it outside.

AdriVanHoudt:
ok so the livereload doesn't do it :P

AdriVanHoudt:
@nlf can you verify #210 ? I want to do a PR with the tweaks I am able to do:P

AdriVanHoudt:
`API rate limit exceeded` :/

PR #209, opened 2015-07-28
add hapi-redis to 3rd party plugins list

sandfox:


Marsup:
@nlf I think it's high time we create a Database section ;)

nlf:
it sounds like you may be right

PR #176, opened 2015-04-24
Updated all dependencies

AdriVanHoudt:
For some reason using the `building-static-server` errors on my machine with 
```js
$ npm start

> hapijs.com@1.0.0 start c:\Users\adri\Documents\GitHub\HapiJS\hapijs.com
> node server.js

  bss:info Livereload ready +0ms
  bss:info Build queued +3ms
  bss:info Build started +302ms
events.js:85
      throw er; // Unhandled 'error' event
            ^
Error: spawn npm ENOENT
    at exports._errnoException (util.js:746:11)
    at Process.ChildProcess._handle.onexit (child_process.js:1053:32)
    at child_process.js:1144:20
    at process._tickDomainCallback (node.js:381:11)

npm ERR! Windows_NT 6.3.9600
npm ERR! argv "c:\\Program Files\\nodejs\\node.exe" "c:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "start"
npm ERR! node v0.12.2
npm ERR! npm  v2.7.4
npm ERR! code ELIFECYCLE
npm ERR! hapijs.com@1.0.0 start: `node server.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the hapijs.com@1.0.0 start script 'node server.js'.
npm ERR! This is most likely a problem with the hapijs.com package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node server.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls hapijs.com
npm ERR! There is likely additional logging output above.
```

Without it it works fine.

PR #211, opened 2015-07-28
updated all the dependencies

AdriVanHoudt:


AdriVanHoudt:
continuation of https://github.com/hapijs/hapijs.com/pull/176

AdriVanHoudt:
@nlf do you want me to just remove `building-static-server`? I would just recommend nodemon or something but that doesn't re-make thought

nlf:
if it's working now, let's leave it. i think we'll eventually replace it with webpack or something though

AdriVanHoudt:
idd, probably was a windows issue, also what is the reason for the special build for hapi? looks like v3 is even with your hapi8 branch?

nlf:
it hadn't been published at the time, go ahead and change it to the regular release

AdriVanHoudt:
:+1:

PR #212, opened 2015-07-29
centered the updates // allow to show partial updates with limit query

### LINKED ISSUE #210: changelog tweaking
* center changelog but keep it left aligned
* only show last 5-10 versions (maybe load more button?)

see https://github.com/hapijs/hapijs.com/pull/207#issuecomment-125682108

AdriVanHoudt:
This will allow to specify how many milestones you want to see (default 10) with the limit query, the `show all` button shows 100 which is probably all :D also this works really nice because they are all cached anyway so the reload is practically instant.

closes #210


AdriVanHoudt:
preview with 1 milestone

mtharrison:
Maybe it should go back to 100% at smaller screen widths, otherwise it gets a bit weird:

screen shot 2015-07-29 at 11 58 58

mtharrison:
Does it make sense to centre the Changelog 
screen shot 2015-07-29 at 12 09 42
title too?

AdriVanHoudt:
fixed the stylings and yeah I could up the limit to 1000 or something but the limit is set to 100 on the fetching.
Centering the changelog definitely makes sense to me and yes I can reset the width on smaller screens.

* up limit (?)
* center title
* reset width of container on small screens

nlf:
i think 100 in the limit seems perfectly reasonable. if someone wants to go back further than that, then they have bigger problems than not having a nice display of the changelog.

i agree with centering the title and resetting the width on small screens, though.

nlf:
i'm curious if it would look better with the version number centered too..

AdriVanHoudt:
I will play around with it a bit later

AdriVanHoudt:
title center looks good 
milestone center looks weird 
I also changed it so that this doesn't happen and the style doesn't change on mobile

AdriVanHoudt:
anything that needs to be done on this?

AdriVanHoudt:
rebased on master

PR #216, opened 2015-08-07
Update plugins tutorial

binarytales:
I have made a small suggested change to the plugins tutorial that clarifies the different ways the plugins can be registered. It was something I was not clear on at first and had to figure out a little myself.

I think this change makes it clearer that the simplest way to register a plugin is to pass register "a plugin registration function". Then progresses to an array of registration functions then expands that you can pass an object with register and options properties and then finally an array of these objects.

It seems also that this was the original intention of the tutorial.

I have only just started developing with Hapi in the last few days so I apologise if I have misunderstood something about the intention here.

I also note that you can mix functions and objects in the array but this is probably an uanessacry complication to add to this tutorial so I have not added an example of this.

### DIFF
--- a/lib/tutorials/plugins.md
+++ b/lib/tutorials/plugins.md
@@ -86,26 +86,18 @@ The return value of `server.select()` is a server object that contains only the 
 
 ## Loading a plugin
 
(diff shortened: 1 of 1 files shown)


devinivy:
Looks like a nice improvement to me.

PR #218, opened 2015-08-09
add book on hapi to resources list

thegoleffect:


hueniverse:
I would add it in second place for now since there is already a book in there.

danielb2:
Does it make sense to add publication dates? May be useful down the line for users to see which ones are more recent etc.

hueniverse:
It's a tricky thing with early previews, electronic editions, etc. I have no interest in starting a debate about which book gets listed first based on some external date because it's just going to end up badly.

This list should be first come first listed basis. When it gets a bit longer I would recommend random listing on the site.

danielb2:
My question was regarding adding it to the PR itself as opposed to specifically suggesting that the order would be impacted by the publication dates. 

First come first listed is fine with me. I'm thinking from the perspective of the interested reader. A book being listed first doesn't mean it's the oldest. There may be a second edition etc. Publication dates would solve that.

hueniverse:
This is not the place to resolve this. People should look into each book and figure out which one is the best for them.

thegoleffect:
updated with books sorted on date added to list

PR #219, opened 2015-08-10
add hapi-info

danielb2:
hapi-info adds route to show hapi version and plugin version

Added to utility section


hueniverse:
So basically, you re-implemented furball :-)

danielb2:
I was worried for a moment I'd missed a plugin for a moment :)

danielb2:
If it makes more sense, we can revive furball instead?

hueniverse:
I don't care.

nlf:
i'm gonna go ahead and merge this, but yeah maybe it would be best to revive furball and then take this down later..

PR #222, opened 2015-08-12
Sort fetched milestones using semver module rather than by milestone …

### LINKED ISSUE #221: Updates page shows milestones in incorrect order
Milestone 9.0.0 is lower down on the page than 8.8.1 because it has a lower ID number (it was created before). I will address this in a separate PR by using the node semver module to sort the milestones.

screen shot 2015-08-12 at 08 57 31

mtharrison:
…id. Closes #221

When sorting by milestone id, version numbers may be out of date.

### DIFF
--- a/lib/changelog.js
+++ b/lib/changelog.js
@@ -1,5 +1,6 @@
 var Async = require('async');
 var Config = require('getconfig');
+var Semver = require('semver');
 var Utils = require('./utils');
 var Querystring = require('querystring');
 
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
:shipit:

lib/changelog.js
```diff
@@ -38,10 +39,10 @@ exports.methods.push({
 
             var compare = function (a, b) {
 
-                if (a.id < b.id) {
+                if (Semver.lt(a.title, b.title)) {
```
Marsup: You're reimplementing Semver.compare.

mtharrison: @Marsup Thanks, I didn't know that existed.

Issue #223, opened 2015-08-12
Some sub-updates don't show


AdriVanHoudt:
see hapijs.com/updates the 9.0 milestone 
![](http://puu.sh/jyAz0/ab7a904e46.png)
![](http://puu.sh/jyAzD/3dc019d5d1.png)

mtharrison:
The content one is because 1.0.2 is still open: https://github.com/hapijs/content/milestones
The subtext one is because it should be `to x.x.x from x.x.x`

We could address both issues maybe by:

 - Include still open milestones for deps (maybe we shouldn't do this though. If it's been released I'm guessing milestone should be closed)
 - Expand the regex to capture this pattern if it's a common one

AdriVanHoudt:
opened https://github.com/hapijs/content/issues/6
and I guess it would be nice to allow `from ... to ...` and `to ... from ...` seems logical

mtharrison:
I updated the regex in a PR
![screen shot 2015-08-12 at 10 00 51](https://cloud.githubusercontent.com/assets/916064/9220258/12848a42-40d9-11e5-92ee-c69d6381dcc6.png)

AdriVanHoudt:
:+1:

AdriVanHoudt:
Content should be fixed now, milestone was closed

hueniverse:
Is this resolved?

AdriVanHoudt:
#224 should have yes (not tested)

hueniverse:
Please close when confirmed.

Marsup:
I've redeployed the site, is it fixed ?

mtharrison:
Looks fixed. Thanks.

PR #224, opened 2015-08-12
Update dependency regex

mtharrison:
Updated dependency regex  to allow either "from x to y" or "to y from x….

See issue #223 for discussion.


AdriVanHoudt:
:shipit: :D

PR #229, opened 2015-08-12
Closes #228

### LINKED ISSUE #228: fix homepage example with hapi 9.x.x

lloydbenson:


lloydbenson:
Updated the README quick as this threw me off when I was testing.  It wasn't clear if I needed to add the files in public that get generated via the make too, if so let me know and I can add those as well.

PR #235, opened 2015-08-24
Update Views tutorial to reflect v9.0.0 changes

jonhester:
Just ran into my first 9.0.0 issue so figured I'd update it for everyone else.


Marsup:
The server.views should be used in the callback, not immediately after like it's synchronous.

hueniverse:
Well, in this case it's fine. This is how hapi used to do it in 8.x. I agree it would be better to use the best common practice, but it is valid at this point.

PR #236, opened 2015-08-24
Update Views tutorial to move server.views into callback

jonhester:
Per discussion in https://github.com/hapijs/hapijs.com/pull/235. I followed the examples from https://github.com/hapijs/vision/blob/master/README.md. I also updated the other full example in the tutorial so I could add the compatibility notice to the top.

### DIFF
--- a/lib/tutorials/views.md
+++ b/lib/tutorials/views.md
@@ -1,5 +1,7 @@
 ## Views
 
+_This tutorial is compatible with hapi v9.x.x._
+
 hapi has extensive support for template rendering, including the ability to load and leverage multiple templating engines, partials, helpers (functions used in templates to manipulate data), and layouts.
 
 ## Configuring the server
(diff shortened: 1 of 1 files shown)


lib/tutorials/views.md
```diff
@@ -231,21 +235,28 @@ server.connection({
...
+server.register(require('vision'), function (err) {
 
-server.route({
-  method: "GET",
-  path: "/",
-  handler: function(request, reply) {
-    reply.view("index");
-  }
+    if (err) {
+        throw err;
```
hueniverse: Generally, `Hoek.assert(!err, err);` is nicer.

jonhester:
@hueniverse Wow, it really is. Now I've learned two new things just from updating this page that I wouldn't have known otherwise. Thanks!

nlf:
nice, thanks!

Issue #215, opened 2015-08-01
Can't click on some footer links


mtharrison:
The "Twitter" and "Need Help" links in the footer can't be clicked because the fixed position sidebar overlaps them. Think a `position:relative` on the footer should fix it.

![screen shot 2015-08-01 at 15 11 23](https://cloud.githubusercontent.com/assets/916064/9022117/9b4e7a36-385f-11e5-8322-e1e859e1d41d.png)

joshvoll:
let me fix this for you @mtharrison

PR #237, opened 2015-08-24
fixes issue #215 Can't click on some footer links

joshvoll:
fixing the footer links
modified file: /css/sections/footer.styl

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -11,8 +11,6 @@ exports.download = function (url, options, callback) {
     Wreck.get(url, options, function (err, res, payload) {
 
         if (err || res.statusCode !== 200) {
-            console.log(res.statusCode);
-            console.log(payload);
             return callback(err || new Error('ERROR: Failed to download, got statusCode: ' + res.statusCode));
         }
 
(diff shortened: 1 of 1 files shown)


Marsup:
Way too many changes on modes. Also this is not the correct solution for having tested this myself, this needs a much deeper reorganization of the elements. While this method fixes it apparently, the sidebar is still overflowing on the bottom if you unfold it enough, which happens very easily in the API section.

joshvoll:
thanks for the comment, let me make other improvements

nlf:
wow, yeah. i'll need you to revert all the file mode change commits for sure. also don't commit the `config/default.json` file, instead create a `config/local.json` file with your real key in it. you're gonna want to regenerate that key asap since it's now public.

also, i agree with @marsup this is a step in the right direction, but it's not quite there yet. good effort so far though!

joshvoll:
Thanks @nlf let me revert all the file change, and keep working on this issue

joshvoll:
@nlf i just reverted the file to the lasted commit, please review and let me know

joshvoll:
Also i just generated a new token for myself, going to work fixing the click issue now

PR #240, opened 2015-08-25
Lint javascript within tutorials.

### LINKED ISSUE #217: "Lint" the tutorials
I think it would be worthwhile to ensure that all the tutorials abide by the hapi style guide.  I see some minor lapses, such as missing new lines at the top of function definitions.  Might feel nice for it to be totally consistent :)

Anyone know if there's a way we can leverage lab to lint the inline code examples?

devinivy:
Resolves #217.

To try it out, `npm run lint:tutorials`.

Three things worth mentioning here,

1. eslint has to be installed locally to the project to take advantage of the shared hapi config, so it's a dev dependency.
2. Object literals alone are not valid javascript, and there's no reasonable way to ask eslint to ignore that.  I generally changed these to [`json5`](http://json5.org/) fenced code blocks, which should cause no display differences.
3. The plugin I used to lint within markdown files ([eslint/eslint-plugin-markdown](https://github.com/eslint/eslint-plugin-markdown)) is not on npm yet and has no tags, so in the package.json I reference a specific git commit from which to install.  https://github.com/eslint/eslint-plugin-markdown/issues/1 lays out the roadmap to get it on to npm.  Sounds like eslint will eventally use it to lint their own docs.


lib/tutorials/cookies.md
```diff
@@ -25,7 +25,7 @@ This configuration will make it so the cookie named `data` has a session time-li
 
 In addition to this, you may pass two parameters to the `config` when adding a route:
 
-```javascript
+```json5
```
AdriVanHoudt: what is the difference?

devinivy: Object literals alone aren't valid javascript– try copying and pasting this code block into a node script / browser console / node console.  You'll get a parse error.  In order to lint, we can't have parse errors, so I changed this fenced code block in a way that should not affect its display while not calling it "javascript" (which it technically isn't).  I wrote briefly about this in the PR.

mtharrison: You could just show the code for adding a route instead? It might be more helpful too than just this floating literal.

nlf:
i tend to agree with @mtharrison that maybe we should just change those object literals to be real code, but that's a separate issue from this in my mind. this is great, quite a few errors and typos it found.

thanks a ton for doing the leg work on this one @devinivy i'm gonna go ahead and merge this as is. we can always create a new issue to expand the object literals to actual code later.

PR #241, opened 2015-08-26
Normalize line endings

maxbeatty:
When I did a fresh clone, I got line ending diffs on OS X with `git config core.autocrlf` set to `input`. [GitHub suggests adding a `.gitattributes` file](https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings) to minimize these differences amongst developers.

What setting should I be using to avoid these diffs?

Does `.gitattributes` need to be more than this generic default?


nlf:
defaults should be fine, it looks like the only issues it had were with a couple of the svg files anyway

Issue #225, opened 2015-08-12
Add inert, h2o2, and vision to plugins


hueniverse:

nlf:
closed via #242

PR #242, opened 2015-08-27
Add inert, h2o2, and vision to plugins

jonhester:
https://github.com/hapijs/hapijs.com/issues/225. I added everything under `The extended hapi universe` since there is some inconsistency as to where first party plugins should go. For example, `good` is under logging while `joi` is not under `Validation`.


nlf:
this is fine for now, but yeah we'll need to move things around eventually. thanks!

PR #244, opened 2015-08-27
make footer position relative so links clickable through sidebar

### LINKED ISSUE #215: Can't click on some footer links

maxbeatty:
attempts to fix #215 

adding `position: relative` makes all links in the footer clickable when there is a sidebar on the page. a good page to verify this fix is `/resources`


Marsup:
Dupe of #237 and still wrong solution, sorry :)

PR #245, opened 2015-08-28
dynamically set height of nav while scrolling

### LINKED ISSUE #215: Can't click on some footer links

maxbeatty:
Another attempt to fix #215 without reorganizing any elements

as you scroll and the nav becomes fixed, its height will now be set to match the remaining height of its partner column (the one with all the content). when you get to the footer, all links will be clickable. if the sidebar overflows, it becomes scrollable so you can still access the content.

### DIFF
--- a/public/css/sections/api.styl
+++ b/public/css/sections/api.styl
@@ -14,7 +14,7 @@
     height: 100%
 
     nav
-      height: 100%
+      height: calc(100% - 135px)
       overflow: auto
     
       & > ul
--- a/public/css/sections/api.styl
+++ b/public/css/sections/api.styl
@@ -14,7 +14,7 @@
     height: 100%
 
     nav
-      height: calc(100% - 135px)
+      height: 100%
       overflow: auto
     
       & > ul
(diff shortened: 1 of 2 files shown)


Marsup:
That'll probably work but I think a better solution would be to let the browser do its job and rearrange elements so that `clear: both` can actually work. Did you try that ?

maxbeatty:
My CSS is a little rusty but I believe that since the sidebar becomes `fixed` on scroll, it is taken out of the normal flow where it can be cleared. [The `height: 100%` rule on `.page-api .nav-secondary`](https://github.com/hapijs/hapijs.com/blob/master/public/css/sections/api.styl#L14) is causing the fixed sidebar to overlap the footer in the normal content flow. Dynamically adjusting the height so it doesn't cover the footer was the solution with the least amount of changes to everything else.

I could try changing the position of the nav to be `absolute` within the sidebar parent and dynamically adjust its `top` position, but that feels like reimplementing what `position: fixed` does in exchange for the layout wrapping as it should. Hopefully I didn't butcher all that CSS lingo too badly. Was that the approach you had in mind? Something different altogether?

nlf:
this is a decent solution, so i'm gonna go ahead and merge it. it gets the job done

Issue #264, opened 2015-09-25
Getting Started could ramp up slowly with more narrative


strand:
I was reading over the [getting started](http://hapijs.com/tutorials/getting-started) guide last night and was surprised when the guide jumped from routes to using a plugin for logging. I am completely new to Hapi, and I expected there to be some progression like (strikethrough indicates we have content for this):
- ~~hello world~~
- ~~displaying a static page~~
- displaying a dynamic page (based on route)
- implementing app reloading in development
- adding a persistence layer
- displaying a dynamic page (based on persistence)
- adding logging
- ~~other plugins~~

Would the Hapi team be amenable to this sort of documentation improvement? I'd be glad to write documentation to ensure that new-comers can more easily dive in.

lloydbenson:
I suspect folks would be most welcome to any ideas here on improvement.  Maybe do a PR with a brief outline for feedback?

strand:
I suspect this will be several PRs, so I'll see if I can write a section about displaying a static page in the next day or two.

nlf:
That sounds like a perfect approach :)

strand:
@hueniverse, thanks for closing. Obviously, this fell off my plate.

PR #265, opened 2015-09-26
Create a section on static pages in Getting Started

strand:
This addresses issue #264 

I am new to Hapi, and would appreciate guidance to ensure this documentation meets standards while introducing jargon and maintaining the right tone and assumed level of competency.

### DIFF
--- a/lib/tutorials/getting-started.md
+++ b/lib/tutorials/getting-started.md
@@ -71,35 +71,39 @@ The `method` parameter can be any valid HTTP method, array of HTTP methods, or a
 
 We've proven that we can start a simple Hapi app with our Hello World application. Next, we'll use a plugin called **inert** to serve a static page. Before you begin, stop the server with **CTRL + C**.
 
-To install **inert** run this command at the command line: `npm install --save inert` This will download inert and add it to your `package.json`, which documents which packages are installed.
+To install inert run this command at the command line: `npm install --save inert` This will download inert and add it to your `package.json`, which documents which packages are installed.
 
 Add the following to your `server.js` file:
 
 ``` javascript
 server.register(require('inert'), function (err) {
-    if (err) { throw err; }
+    if (err) {
+        throw err;
+    }
+
+    server.route({
+        method: 'GET',
+        path: '/hello',
+        handler: function (request, reply) {
+            reply.file('./public/hello.html');
+        }
+    });
 });
 
-server.route({
-    method: 'GET',
-    path: '/hello',
-    handler: function (request, reply) {
-        reply.file('./public/hello.html');
-    }
-});
+
 ```
 
-The `server.register` command above adds the inert plugin to your Hapi application. If something goes wrong, we want to know, so we've passed in an anonymous function which if invoked will receive `err` and throw that error. `err` is used instead of `error` because `error` is a reserved word in JavaScript. This callback function is required when registering plugins.
+The `server.register()` command above adds the inert plugin to your Hapi application. If something goes wrong, we want to know, so we've passed in an anonymous function which if invoked will receive `err` and throw that error. `err` is used instead of `error` because `error` is a reserved word in JavaScript. This callback function is required when registering plugins.
 
-The `server.route` command registers the `/hello` route, which tells your server to accept GET requests to `/hello` and reply with the contents of the `hello.html` file.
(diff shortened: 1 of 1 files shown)


lib/tutorials/getting-started.md
```diff
@@ -67,6 +67,40 @@ Note that we URI encode the name parameter, this is to prevent content injection
...
+To install **inert** run this command at the command line: `npm install --save inert` This will download inert and add it to your `package.json`, which documents which packages are installed.
+
+Add the following to your `server.js` file:
+
+``` javascript
+server.register(require('inert'), function (err) {
+    if (err) { throw err; }
+});
+
+server.route({
```
nlf: move this `server.route` call inside the callback for `server.register` above

lib/tutorials/getting-started.md
```diff
@@ -67,6 +67,40 @@ Note that we URI encode the name parameter, this is to prevent content injection
...
+});
+```
+
+The `server.register` command above adds the inert plugin to your Hapi application. If something goes wrong, we want to know, so we've passed in an anonymous function which if invoked will receive `err` and throw that error. `err` is used instead of `error` because `error` is a reserved word in JavaScript. This callback function is required when registering plugins.
+
+The `server.route` command registers the `/hello` route, which tells your server to accept GET requests to `/hello` and reply with the contents of the `hello.html` file.
+
+Start up your server with `npm start` and go to `http://localhost:3000/hello` in your browser. Oh no! We're getting a 404 error because we never created a `hello.html` file. You need to create the missing file to get rid of this error.
+
+Create a folder called `public` at the root of your directory with a file called `hello.html` within it. Inside `hello.html` put the following HTML: `<h2> Hella Whirled.</h2>`. Then reload the page in your browser. You should see a header reading "Hella Whirled."
```
nlf: let's use "Hello World" instead of being creative, for simplicity :)

lib/tutorials/getting-started.md
```diff
@@ -67,6 +67,40 @@ Note that we URI encode the name parameter, this is to prevent content injection
...
+
+Inert will serve whatever content is saved to your hard drive when the request is made, which is what leads to this apparent live reloading behavior. Customize the page at `/hello` to your liking.
+
+More details on how static content is served are detailed on [Serving Static Content](http://localhost:3000/tutorials/serving-files). This techniques is commonly used to serve images, stylesheets, and static pages in your web application.
```
nlf: make this a relative link, `/tutorials/serving-files`

nlf:
added some notes, but this is a great start!

strand:
Thanks for the feedback, I'll integrate it as soon as I have time for hobby coding (probably over the weekend).

lib/tutorials/getting-started.md
```diff
@@ -67,6 +67,40 @@ Note that we URI encode the name parameter, this is to prevent content injection
...
+
+We've proven that we can start a simple Hapi app with our Hello World application. Next, we'll use a plugin called **inert** to serve a static page. Before you begin, stop the server with **CTRL + C**.
+
+To install **inert** run this command at the command line: `npm install --save inert` This will download inert and add it to your `package.json`, which documents which packages are installed.
+
+Add the following to your `server.js` file:
+
+``` javascript
+server.register(require('inert'), function (err) {
+    if (err) { throw err; }
```
nlf: almost forgot, this should be separate lines

```js
if (err) {
    throw err;
}
```

lib/tutorials/getting-started.md
strand: As far as I can tell, `server.route` works both inside and outside of `server.register`. Is having route inside register a commonly-used convention in Hapi.js? Why is it preferable to declaring the route at the same level as the register command?

lib/tutorials/getting-started.md
strand: D'oh!

lib/tutorials/getting-started.md
nlf: the idea with calling `server.route` in the callback is to demonstrate that it's generally wise to run code that depends on a plugin within the callback to register so that you can be absolutely sure that plugin exists when your code runs.

hueniverse:
Function references should include `()`

strand:
@hueniverse Thank you! I write Ruby most of the time, so I've grown accustomed to omitting parens. WILL_FIX

strand:
Sorry for the slow turnaround. I think this is really close, and my next step will be to start working on displaying dynamic html, as listed in #264.

nlf:
this looks great! good job! i think we're ready to merge this, unless you want me to wait? you can always submit a new pull request to add more sections to the tutorial later.

strand:
Merge it! I jope to get the next section drafted in the next couple of days.

PR #266, opened 2015-09-30
Adding Hoek dependency to views tutorial

ryanmurakami:
The Hoek import in the views tutorial is missing. Adding it in so the code works with no errors.


devinivy:
Looks right to me.

PR #280, opened 2015-10-28
shorten sidebar height so the pages feel less cramped

franvarney:
screen shot 2015-10-28 at 1 52 21 pm


nlf:
at this height does the nav for the api docs feel big enough? i almost wonder if this should also make the search box and reference version dropdown smaller.

also, does this mean we could remove the code that shifts the footer over on the api docs?

franvarney:
I don't personally think it's too small? But I could see why some people might think it is. I open to making the search box/version dropdown smaller. The search input box might be able to be inlined with the 'search' text even.

And maybe. Need to check more resolutions first to make sure.
screen shot 2015-10-28 at 2 13 17 pm

PR #282, opened 2015-10-29
add hapi-alive plugin

idosh:


nlf:
can you change this to lowercase the H in hapi.js? the brand is always lowercased.

nlf:
thanks!

PR #289, opened 2015-11-04
Update the homepage example server

johnbrett:
Updated the homepage example to use es6 and hapijs styleguide

### DIFF
--- a/templates/index.jade
+++ b/templates/index.jade
@@ -153,10 +153,16 @@ block content
           span.pl-c // Start the server
           br
 
-          | server.start(() => {
+          | server.start((err) => {
           br
           
           br
+          span.pl-s     if 
+          | (err) {
+          br
+          span.pl-s         throw 
+          | err;
+          |     }
           |     console.log(
           span.pl-s1
             span.pl-pds '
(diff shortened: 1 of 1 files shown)


templates/index.jade
```diff
@@ -142,8 +153,16 @@ block content
           span.pl-c // Start the server
           br
 
-          | server.start(function () {
-          |     console.log('Server running at:', server.info.uri);
+          | server.start(() => {
```
AdriVanHoudt: server.start returns an error (if there is one), maybe also deal with that

templates/index.jade
```diff
@@ -142,8 +153,22 @@ block content
...
+          | server.start((err) => {
+          br
+          
+          br
+          span.pl-s     if 
+          | (err) {
+          br
+          span.pl-s         throw 
+          | err;
+          |     }
```
johnbrett: @AdriVanHoudt that better? :)

AdriVanHoudt: that jade is not super readable but looks good :+1:

nlf:
:+1: thanks!

PR #303, opened 2015-12-11
Add hapi-recaptcha plugin to the list

cristobal151:
A simple google's recaptcha plugin for hapi.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -389,6 +389,10 @@ exports.categories = {
         tv: {
             url: 'https://github.com/hapijs/tv',
             description: 'Interactive debug console plugin for hapi servers'
+        },
+        'hapi-recaptcha': {
+            url: 'https://github.com/cristobal15/hapi-recaptcha',
+            description: 'Google's reCaptcha for hapi'
         }
     },
     Validation: {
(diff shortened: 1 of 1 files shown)


lib/plugins.js
```diff
@@ -389,6 +389,10 @@ exports.categories = {
         tv: {
             url: 'https://github.com/hapijs/tv',
             description: 'Interactive debug console plugin for hapi servers'
+        },
+        'hapi-recaptcha': {
+            url: 'https://github.com/cristobal15/hapi-recaptcha',
+            description: 'Google's reCaptcha for hapi'
```
arb: This isn't valid JS.

nlf: what he means by that is you need to escape the single quote character like `Google\'s`

lib/plugins.js
```diff
@@ -389,6 +389,10 @@ exports.categories = {
         tv: {
             url: 'https://github.com/hapijs/tv',
             description: 'Interactive debug console plugin for hapi servers'
+        },
+        'hapi-recaptcha': {
+            url: 'https://github.com/cristobal15/hapi-recaptcha',
```
gergoerdosi: The URL isn't valid either, your username is wrong.

cristobal151:
Sorry for the mistakes.
I already made changes in the url and description.

Thanks.

PR #304, opened 2015-12-11
My views experience

zigomir:
I needed to install `handlebars` myself.

`Path` was not needed at all in this step for me.


zigomir:
Ah, I see the need for `Path` later on for layouts. Will prepare another fix. But `handlebars` still needs to be installed first in my experience...

Marsup:
As we didn't explain how to install vision, we didn't either for handlebars, that's assuming you have basic knowledge about node at all, you shouldn't start creating a server if you don't even know how to install modules.

PR #309, opened 2015-12-22
Update plugins.js

toymachiner62:
Added hapi-authorization

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -3,6 +3,10 @@ exports.categories = {
         'hapi-rbac': {
             url: 'https://github.com/franciscogouveia/hapi-rbac',
             description: 'A Rule Based Access Control module for hapi.'
+        },
+        'hapi-authorization': {
+            url: 'https://github.com/toymachiner62/hapi-authorization',
+            description: 'ACL Support for hapi apps.'
         }
     },
     'Authentication': {


mtharrison:
Just out of curiosity, what does this offer over the built-in scopes functionality?

toymachiner62:
Idk. I'm not familiar with the scopes functionality. Can you point me to some good documentation?

mtharrison:
You can find `scope` under [route options](http://hapijs.com/api#route-options).

There's also this blog post: https://blog.andyet.com/2015/06/16/harnessing-hapi-scopes/ and an SO answer I wrote a while back http://stackoverflow.com/a/29053887/1402929.

toymachiner62:
Without diving too deeply into scopes, one thing that it provides is hierarchical authorizations. 

Do scopes allow you to blacklist all routes so that they require a certain role without having to specify on every route? Like this: https://www.npmjs.com/package/hapi-authorization#blacklist-all-routes-to-require-authorization

nlf:
You can set a default required scope for every route, which effectively does the same thing.

toymachiner62:
Interesting. Well maybe hapi-authorization is obsolete then. 

@AdrieanKhisbe @asafdav do you guys know of any other differences between hapi-authorization/hapi-auth-extra that the built in hapi scope does not provide?

asafdav:
Hi @toymachiner62 I think that the main difference is ACL support

PR #313, opened 2016-01-01
plugins update: hapi-marko

pvde:


Marsup:
I'd question referencing such plugins that are bad citizens, there's a well established view architecture in place (vision) it should integrate into that, makes for a much simpler life for end users.

nlf:
that's a valid concern.. is there a reason this module doesn't just integrate with vision?

pvde:
Ok thank you for the feedback. I will close this pull request for now since i'm not the maintainer of this plugin. @Marsup: Is there a page somewhere with guidelines on how plugins should be developed?

Marsup:
No but this one was kind of obvious considering how all of the other plugins did it.

PR #323, opened 2016-01-15
Add the `therealyou` plugin

briandela:
Add the `therealyou` plugin to the http://hapijs.com/plugins page.


nlf:
nice, very handy

PR #324, opened 2016-01-19
Plugin: hapi-sequelize-crud

mdibaiee:


Marsup:
FWIW I think your readme will be a bit unsettling for people not yet using ES7 ;)

PR #334, opened 2016-02-11
Update server-methods.md

catalint:
https://github.com/hapijs/catbox#policy states that expiresAt is in `HH:MM` format, also, updated other help texts with new info from catbox/readme.md


Marsup:
LGTM. Maybe we should additionally link to catbox docs.

PR #336, opened 2016-02-16
Add 'hapi-mongoose' plugin

asilluron:
Add the `hapi-mongoose` plugin to the list of plugins.


Marsup:
FWIW I don't think server methods is the correct way to expose your API, you'd better use expose or decorate.

asilluron:
<<updated>>

It's worth a lot. 

Thanks for the feedback.

PR #338, opened 2016-02-23
add testing-hapi boilerplate link

pashariger:


nlf:
i feel like it's important that you mention somewhere in your readme that your eslint configuration changes are incompatible with the hapi style guide, just to make it clear to anyone using your boilerplate that that's the case. especially since the few changes you have make a pretty dramatic difference in how the code looks.

pashariger:
Good point. I updated the code to conform to the hapi style guide and eslint rules, which should be friendlier to anyone reading through it.

nlf:
perfect! thanks!

PR #340, opened 2016-02-25
Fix views tutorial to put views only in "./templates"

### LINKED ISSUE #333: Fix directory structure consistency in Views tutorial
There are a few parts that you use a `templates` base directory, as in the first code block:

```javascript
    server.views({
        engines: {
            html: require('handlebars')
        },
        relativeTo: __dirname,
        path: 'templates'
    });
```

In other parts, you use `views` as a base directory:

```
For example, if you have a directory structure like:

views\
  index.html
  layout\
    layout.html
  helpers\
    getUser.js
    fortune.js
```

Considering that you use `templates` as a base directory in this repository, I would guess that's what you recommend. I'd be happy to send you a pull request if that's the case :+1:

vitorbaptista:
We were inconsistent, using both "./views" and "./templates". This changes the
docs to use only "./templates".

Fixes #333.

### DIFF
--- a/lib/tutorials/views.md
+++ b/lib/tutorials/views.md
@@ -91,7 +91,7 @@ Additionally, there are two settings that alter how hapi will allow you to use p
 For example, if you have a directory structure like:
 
 ```
-views\
+templates\
   index.html
   layout\
     layout.html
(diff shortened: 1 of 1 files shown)


nlf:
:+1: thanks

Issue #325, opened 2016-01-20
Upgrade eslint-plugin-markdown to v1


devinivy:
We currently install eslint-plugin-markdown from github.  Now it's published as v1-beta on npm.

nlf:
PR?

devinivy:
PR #341 posted.  Sorry about the delay– I was sort of thinking it would make its way out of beta quickly, but it didn't!

PR #341, opened 2016-02-26
Use latest beta of eslint-plugin-markdown

devinivy:
See #325.  We'll have to update this again when the eslint plugin leaves beta, but it does currently seem to be working well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "building-static-server": "3.x.x",
     "eslint": "1.0.x",
     "eslint-config-hapi": "1.x.x",
-    "eslint-plugin-markdown": "git://github.com/eslint/eslint-plugin-markdown.git#45f03b5",
+    "eslint-plugin-markdown": ">=1.0.0-beta.1 <2.0.0",
     "glob-cli": "1.x.x",
     "stylus": "0.52.x",
     "uglify-js": "2.x.x"


Marsup:
Isn't it a bit too broad ?

devinivy:
This will include `1.0.0-beta.x` if I understand how npm treats prerelease ranges.  Is that too broad?

Marsup:
It will also include any future version like 2.x.

devinivy:
Really?  From https://docs.npmjs.com/misc/semver#prerelease-tags,
>For example, the range >1.2.3-alpha.3 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by 3.4.5-alpha.9, even though 3.4.5-alpha.9 is technically "greater than" 1.2.3-alpha.3 according to the SemVer sort rules.

Marsup:
It's important to read until the end ;)
> The version range only accepts prerelease tags on the 1.2.3 version. The version 3.4.5 would satisfy the range, because it does not have a prerelease flag, and 3.4.5 is greater than 1.2.3-alpha.7.

devinivy:
Yup, I sure found what I was looking for when I started reading :P

In any case, now it's updated to accept v1.0.0-beta up to (but not including) v2.

nlf:
nice, thanks

Issue #339, opened 2016-02-23
Tools/utilities that don't fit as plugins


danielb2:
I wrote a tool called hapi-runplugin specifically to easily list what endpoints are returned by running a given plugin. When things are organized in different repos and they are re-used and it's not always sensible to test something via a server, this comes in handy.

Regardless, the webpage has a plugins section, but no section for tools. A composer like Glue is a tool like this which also has value.

I've noticed that Glue can easily be overlooked, so this would solve that.

thoughts?

Ref: https://github.com/hapijs/contrib/issues/76

Marsup:
Boilerplates are in the section resources, what would you think about putting tools there too ? Seems similar.

danielb2:
That works. I don't think tools have enough stuff to warrant it's whole own section

danielb2:
I'll just submit a PR to put that in and close this for now.

PR #342, opened 2016-02-27
add tools section to resources with glue and hapi-runplugin

danielb2:
Ref: https://github.com/hapijs/hapijs.com/issues/339

A section for tools such as glue, rejoice and hapi-runplugin that aren't plugins but should not be overlooked in the hapi ecosystem.

### DIFF
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -95,6 +95,20 @@ exports.categories = {
             description: 'A HTML5 multiplayer game, based on the popular board game "Catan" by Klaus Teuber.'
         }
     },
+    'Tools': {
+        'glue': {
+            url: 'https://github.com/hapijs/glue',
+            description: 'Glue provides configuration based composition of hapi\'s Server object.'
+        },
+        'rejoice': {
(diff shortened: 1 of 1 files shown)


Marsup:
Should we leave glue and rejoice in the plugins section then ?

danielb2:
then maybe better to leave the way it is and put runplugin under extended instead?

nlf:
what if we just make an "external tools" section in the plugins page?

the extended universe section is specifically for listing other modules in the hapijs org

danielb2:
Also, maybe the extended universe should be moved to resources since they're not plugins ?

nlf:
my feeling here is still that these tools should stay in the plugins page, though in their own section. putting them in the resources page feels counterintuitive when that page is almost exclusively for documentation

nlf:
ping? i'd like to get these added to the list, but i think it needs to be under a heading on the plugins page instead of its own page

hueniverse:
Closing due to lack of progress.

PR #345, opened 2016-03-06
Add hodgepodge and loveboat to plugins page

devinivy:
The only strangeness here is that hodgepodge is a utility but not a plugin.

Adding,
https://github.com/devinivy/hodgepodge
https://github.com/devinivy/loveboat


Marsup:
Until #342 gets decided...

PR #347, opened 2016-03-08
Add k7 in plugins list

thebergamo:
Maybe is good adding a section for database related stuffs, because we have a big list of database plugins :)


thebergamo:
@nlf some review?

PR #348, opened 2016-03-10
Add request payload utility plugins

charlesbjohnson:
Adding a few plugins to the list that we are using at [Lob](https://lob.com/).

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -262,6 +262,10 @@ exports.categories = {
             url: 'https://github.com/knownasilya/hapi-decorators',
             description: 'Decorators for HapiJS routes'
         },
+        'hapi-default-payload': {
+            url: 'https://github.com/lob/hapi-default-payload',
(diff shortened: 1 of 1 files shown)


nlf:
re: hapi-default-payload why not just make your payload validator `Joi.object().default({})`?

charlesbjohnson:
@nlf good question.

The main reason we use it is because we want the empty object to be present *before* validation takes place. We want Joi validation to fail on both `null` and `{}` (since `{}` is still not a valid input and will just fail somewhere else later in the request lifecycle). The difference is that we want the failure message for `null` to be more similar to the error message that we would get for `{}`.

TL;DR it's a simple workaround addressing a more complicated but missing option from Joi. Ie. more verbose error messages for cases such as...

```js
Joi.validate(null, Joi.object().keys({ a: Joi.integer().required() }))
```

Hope that makes sense.

nlf:
that does make sense actually, thanks for the explanation :)

PR #364, opened 2016-04-08
Add hapi-bookshelf-total-count plugin

elnaz:
Thanks!


nlf:
interesting approach, thanks for the submission!

PR #370, opened 2016-04-26
provide default monospace font

hsablonniere:
On my ubuntu machine, the lack of the default "monospace" CSS font-family gives me code in some serif font :-(

### DIFF
--- a/public/css/globals/_variables.styl
+++ b/public/css/globals/_variables.styl
@@ -5,7 +5,7 @@
 /* Typefaces */
 
 $font-facit = 'jaf-facitweb', 'Helvetica Neue', Helvetica, Arial, sans-serif
-$font-inconsolata = 'inconsolata', menlo, consolas, monospaced
+$font-inconsolata = 'inconsolata', menlo, consolas, monospace
 
 /* Font Sizes */
 


public/css/globals/_variables.styl
```diff
@@ -5,7 +5,7 @@
 /* Typefaces */
 
 $font-facit = 'jaf-facitweb', 'Helvetica Neue', Helvetica, Arial, sans-serif
-$font-inconsolata = 'inconsolata', menlo, consolas, monospaced
+$font-inconsolata = 'inconsolata', menlo, consolas, monospaced, monospace
```
mtharrison: Should `monospaced` be removed? Doesn't seem to be a valid generic family name: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

nlf: yeah i think this was a typo and monospaced should've been monospace from the start

hsablonniere: I was not sure, that's why I added instead of replacing it. It just updated my PR ✌

mtharrison:
LGTM

hsablonniere:
Thx, how does the update on the website works? I cannot see the change yet.

Marsup:
Should be good now, the site needed to be built (makefile).

hsablonniere:
It works!

PR #385, opened 2016-05-24
add disinfect plugin

genediazjr:


lib/plugins.js
```diff
@@ -198,6 +198,10 @@ exports.categories = {
         }
     },
     'Utility': {
+        'disinfect': {
+            url: 'https://github.com/genediazjr/disinfect',
+            description: 'Request query, payload, and params sanitation'
```
Marsup: Meant sanitization here ?

devinivy: I think sanitation and sanitization could both work here.

Marsup: OK, thought it was a typo.

mtharrison: Really? I've never seen "sanitation" used in software terms before.

genediazjr: I agree, I think it would be proper to use sanitization. I'll provide an update.

PR #389, opened 2016-05-27
Added multines to plugins

mcollina:


devinivy:
Excited to try this plugin out some time

mcollina:
Let me know what you think of it! PR always welcomed.

nlf:
very neat! i agree with @devinivy i can't wait to play with this

mcollina:
It should work even with cluster and child-process if used with https://github.com/mcollina/mqemitter-child-process. I need to write some more docs :D.

PR #392, opened 2016-06-03
Add Akaya utility plugins

felixheck:


mtharrison:
It's a common practice (requirement?) to add your name to the license as an addition, rather than removing original author:

http://programmers.stackexchange.com/questions/114701/what-are-the-legal-considerations-when-forking-a-bsd-licensed-project

mtharrison:
I think you need to update URL here:

> or clone the repository:
> 
> // development version with ES6 syntax
> $ git clone https://github.com/mtharrison/akaya

Marsup:
Why add babel when all you're using are node 4 features ?

Marsup:
Can't merge that btw, need to rebase on master.

PR #393, opened 2016-06-07
Packt URL added

tanS30:
Packt reference added.


hueniverse:
This is the wrong place for this link. It should go with the rest of the books.

tanS30:
Can you please let us know where to add the URL, we cannot find the rest of the books section.
Apologies for posting at the wrong place.

mtharrison:
The book is already on the resource page: https://github.com/hapijs/hapijs.com/blob/master/lib/resources.js#L14-L17

hueniverse:
Then that's pretty much it.

PR #402, opened 2016-06-14
Plugins: Add 'hapi-multi-mongo'

metoikos:


Marsup:
Maybe it's time I put my own after 2y, seems we're doing exactly the same thing :)

PR #403, opened 2016-06-14
Add hapi-mongodb

Marsup:


Marsup:
And also fix #402... That linter is seriously useless as it is.

PR #405, opened 2016-06-15
Updated some plugin infoz

danecando:
Deleted abandoned hapi-sequelize and updated it for the current version. Added a new plugin for Rollbar logging service.


Marsup:
The previous owner of that package name should update its readme telling he's discontinuing the project in favor of yours.

danecando:
@Marsup - I realized how it might look after the fact. I've asked them about it before and they told me that they were discontinuing development and gave me the name on npm (mine was previously hapi-sequelized)

Marsup:
I saw that in the diff. I hope at least you started your version number higher than theirs, because people might get an unexpected package if they now have yours.

PR #413, opened 2016-08-02
Add haps-attempts-limiter to plugin list

acavestro:
Hi, 
recently I wrote a plugin for Hapi.js and I would like to add it to Hapi's plugin list.

Feel free to comment my work, any opinion, suggestion or criticism is always well accepted!

Bye
Antonio

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -277,6 +277,10 @@ exports.categories = {
             url: 'https://github.com/ide/hapi-async-handler',
             description: 'Use ES7 async functions and co generator functions as hapi route handlers'
         },
+        'hapi-attempts-limiter': {
+            url: 'https://github.com/acavestro/hapi-attempts-limiter',
+            description: 'An hapi.js plugin that limits the number of failed attempts for every route'
+        },
         'hapi-bookshelf-models': {
             url: 'https://github.com/lob/hapi-bookshelf-models',
             description: 'Load, register, and expose Bookshelf.js models'


AdriVanHoudt:
Did not know about `takeover()`, cool!
Also the console.error seems weird (https://github.com/acavestro/hapi-attempts-limiter/blob/4c62db4ffee25e73e0c8b0c4acc660139270a3d1/lib/index.js#L109), wouldn't it be more useful to do a `request.log` with the `error` tag?

Marsup:
You shouldn't manipulate the headers object directly. I also think you could spare the merge of options on every response but I can't find that example anymore.

acavestro:
@AdriVanHoudt well, I added ```takeover()``` to be sure that the reply is sent immediately, but after another reading of the documentation I'm not so sure about that:

> Note that prerequisites do not follow the same rules of the normal reply interface. In all other cases, calling reply() with or without a value will use the result as the response sent back to the client. In a prerequisite method, calling reply() will assign the returned value to the provided assign key. If the returned value is an error, the failAction setting determines the behavior. To force the return value as the response and ends the request lifecycle, use the reply().takeover() method.

Also, you're right about ```request.log```, I forgot to change those lines in something more appropriate. Thanks for your hint 😄

Marsup:
Just saying you're doing the same job over and over again. applyToDefaults has a cost, however minimal, it's probably possible to avoid it.

PR #419, opened 2016-08-20
Add hapijs-status-monitor to Logging/Metrics section

ziyasal:
plugin [link](https://github.com/ziyasal/hapijs-status-monitor)

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -162,6 +162,10 @@ exports.categories = {
         'hapi-rollbar': {
             url: 'https://github.com/danecando/hapi-rollbar',
             description: 'Hapi plugin for the fabulous Rollbar logging service'
+        },
+        'hapijs-status-monitor':{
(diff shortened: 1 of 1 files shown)


Marsup:
FWIW several comments :
- you should not add methods to JavaScript native types
- you should not directly depend on hapi, it's a plugin, it should be a peer dependency
- your .npmignore could probably also ignore misc and .vscode
- you should use an actual server.route instead of hijacking the onPreResponse
- your response time measurement is wrong, you're measuring the time between 2 statements, explaining the impossible response times of ~0.0010ms

ziyasal:
hi @Marsup thank you for your review 👍 I will refactor it asap (:

Marsup:
Cool plugin btw, I only mentioned what I found wrong but it's a great addition !

ziyasal:
hi @Marsup, thanks! btw, I have made quick refactoring to cover review and pushed new version.

Issue #200, opened 2015-07-17
Improve browsing experience for documentation


kevinsimper:
Sometimes it can be really hard to see where properties is belonging to, for example the `server.route` has a lot of nested configuration, and sometimes you miss that it has to nested a chain of properties.

I have missed that `payload.parse` should be inside the `config` object and not as a property inside the `server.route`.

Accordion
https://jqueryui.com/accordion/#default

mik01aj:
+1 +1 +1. The documentation is very hard to read.

Hapi has a lots of alternative-options-in-configs-in-callbacks-to-I-dont-know-what and this becomes overwhelming when the documentation doesn't have a good visual structure. Here's an example of handler options to the `route` method's `handler` option (afaik normally there's a function there, but you can also have an object):

![screenshot 2015-08-19 09 40 23](https://cloud.githubusercontent.com/assets/3027415/9351088/9e13bf56-4656-11e5-8f15-c66d0e983928.png)

You have your headers in almost the same font face as everything else. The vertical spacing is so big, and the list indent so small that you hardly see which level of nesting you're looking at. (Actually, four levels of nested lists are usually not a good idea, and your docs contain lots of places with 4 levels, and even some with 5 levels).

In short, I mean: some good CSS would help as well. Expanding/collapsing, yes, it's a good idea, but some better styling is needed too. And this should be actually much easier.

It is sooooo easy to get lost.

kevinsimper:
@nlf How can we make this happen?

AdriVanHoudt:
since `app.js` starts with `/* these methods borrowed from youmightnotneedjquery.com */` I assume we need to find a solution without jquery :P

mtharrison:
I agree that some simple styling would help too. Some subtle borders on the `<ul>`s might improve things:

![screen shot 2015-08-19 at 10 13 12](https://cloud.githubusercontent.com/assets/916064/9352727/6065a9bc-465b-11e5-90ba-3ea34c0c78f5.png)

Marsup:
Putting "help wanted" suggests that @nlf is waiting for a PR.

AdriVanHoudt:
@mtharrison nice idea but looks ugly ^^

kevinsimper:
@Marsup But with no information on what they expect and what they think, it is hard to know if the time spent is worth it.

Marsup:
That debate is already a good start, if you all converge on a solution I bet it'll take.

dj:
+1

[...]

PR #422, opened 2016-09-13
Improve API sidebar nav

mtharrison:
Tries to address some of the issues here: https://github.com/hapijs/hapijs.com/issues/200
- Use bootstrap affix plugin to affix the API sidenav
- Use bootstrap scrollspy to show position in doc
- Autoscroll menu

This adds a very trimmed down version of Bootstrap with no CSS and only the Affix and Scrollspy jQuery plugins. There's a bit of custom code to get the menu to autoscroll to current position - this isn't something scrollspy does by itself.

scrollspy

This isn't perfect but I think it's a lot better than what we have now. Welcome feedback/improvements.


devinivy:
This is good news!

nlf:
we used to have the nav follow the scroll position, but we removed it and i can't seem to find the issue that explained why right now.. i personally like the idea, but i wish i could remember why we took it out.

do we really need to bring in both jquery and bootstrap just for this?

also you've got quite a few lines in the diff that are removing whitespace that should be there. we don't have linting of the client side code right now, but i do try to adhere to the hapi style guide anyway

mtharrison:
> we used to have the nav follow the scroll position, but we removed it and i can't seem to find the issue that explained why right now.. i personally like the idea, but i wish i could remember why we took it out.

Do you mean the menu was just fixed position or the actual scroll position of the menu tracked the scroll position of the body too? I can only find commits that set it to fixed and change the menu height (e72b81e and reverted in f68104e).

> do we really need to bring in both jquery and bootstrap just for this?

It's bringing in a tiny bit of bootstrap (6k) with only the 2 plugins. But yes they're jQuery plugins so it requires jQuery (86k). If you're worried about that though, I could look into if there's an alternative lib or if it's something we can write ourselves easily enough. 

> also you've got quite a few lines in the diff that are removing whitespace that should be there. we don't have linting of the client side code right now, but i do try to adhere to the hapi style guide anyway

I'll fix this. Think my editor changed it by itself.

mtharrison:
Added help wanted because I'm unlikely to find any time to improve this soon.

PR #426, opened 2016-10-04
adds information on how to clear a cookie

ferrao:
It took me a while to get to `response.unset()` as the way to erase a cookie while trying to implement a simple logout handler, adding this info to the cookie tutorial should make it easier for others to find.

Both English and Portuguese translation provided.

### DIFF
--- a/lib/tutorials/en_US/cookies.md
+++ b/lib/tutorials/en_US/cookies.md
@@ -50,7 +50,7 @@ In this example, hapi will reply with the string `Hello` as well as set a cookie
 The cookie can be cleared by calling the `unstate()` method on the [`response`](/api#response-object) object:
 
 ```javascript
-reply('Hello').unstate('data', { firstVisit: false });
+reply('Hello').unstate('data');
 ```
 
 ### Overriding options
@@ -59,6 +59,3 @@ When setting a cookie, you may also pass the same options available to `server.s
 
 ```javascript
 reply('Hello').state('data', 'test', { encoding: 'none' });
-```
-
-When clearing a cookie, the specified keys in options must match the same keys in the server definition.  
--- a/lib/tutorials/en_US/cookies.md
+++ b/lib/tutorials/en_US/cookies.md
@@ -50,7 +50,7 @@ In this example, hapi will reply with the string `Hello` as well as set a cookie
 The cookie can be cleared by calling the `unstate()` method on the [`response`](/api#response-object) object:
 
 ```javascript
-reply('Hello').unstate('data', { firstVisit: false});
+reply('Hello').unstate('data', { firstVisit: false });
 ```
 
 ### Overriding options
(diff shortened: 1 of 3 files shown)


ferrao:
It seems i pushed the button to quick, build is breaking due to style issues, will fix and submit a new PR.

lib/tutorials/en_US/cookies.md
```diff
@@ -46,10 +46,19 @@ reply('Hello').state('data', { firstVisit: false });
 
 In this example, hapi will reply with the string `Hello` as well as set a cookie named `data` to a base64 encoded string representation of the given object.
 
+## Clearing a cookie
+The cookie can be cleared by calling the `unstate()` method on the [`response`](/api#response-object) object:
+
+```javascript
+reply('Hello').unstate('data', { firstVisit: false });
```
devinivy: Is this intended to include `{ firstVisit: false }`?

lib/tutorials/en_US/cookies.md
```diff
@@ -46,10 +46,19 @@ reply('Hello').state('data', { firstVisit: false });
...
+
 ### Overriding options
 
 When setting a cookie, you may also pass the same options available to `server.state()` as a third parameter, such as:
 
 ```javascript
 reply('Hello').state('data', 'test', { encoding: 'none' });
 ```
+
+When clearing a cookie, the specified keys in options must match the same keys in the server definition.
```
devinivy: I'm not sure if this is entirely correct, but I am also a little confused by the `unstate()` options.  The official wording in docs is,
>If the state was previously registered with the server using `server.state()`, the specified keys in `options` override those same keys in the server definition (but not others).

lib/tutorials/en_US/cookies.md
ferrao: `firstVisit` is just an example of some previous state. I used this property name to match what was already there on the Setting a Cookie section of the tutorial.

lib/tutorials/en_US/cookies.md
ferrao: On my experiments, ommiting them would cause the cookie to remain in the browser.

lib/tutorials/en_US/cookies.md
AdriVanHoudt: Will doing just `.unstate()` clear the cookie?

AdriVanHoudt: Will doing just `.unstate()` clear the cookie?

ferrao:
@devinivy @AdriVanHoudt just tested this to be sure. As long as you specify the correct cookie name, `.unstate()` will clear it. Have pushed an update to remove the incorrect information.

ferrao:
@devinivy Do you think the information i have added is now accurate?

devinivy:
Yes!  Looks good to me.

Marsup:
`Overriding options` is a subsection of `Setting a cookie`, your new section shouldn't be there.

ferrao:
You are correct @Marsup, i moved it to the `Setting a cookie` section now.

Marsup:
Thanks ! Sorry for the delay.

ferrao:
No prob, i am a very happy hapi user, just glad to contribute with something, even if extremely small.

Marsup:
This is no small contribution believe me :)

PR #427, opened 2016-10-07
Set host to 'localhost' in server.connection

nongaap:
Setting host to 'localhost' ensures http://localhost:3000 console logs and matches tutorial.

Server.info.uri defaults to the operating system hostname and not 'localhost'. I was confused when going through the tutorial and was console.logging my computer name instead of http://localhost:3000 as specified in the tutorial. When I set host to 'localhost' it fixed the issue.


nlf:
good idea!

PR #431, opened 2016-10-19
style fix

danielb2:
* [callbacks should be made with explicit returns](https://github.com/hapijs/contrib/blob/master/Style.md#callback)


Marsup:
I'm not sure I would consider reply a callback.

danielb2:
What is it then? Also, the doc is inconsistent with some of them using an explicit return statement.

danielb2:
I suppose you're right. it's not actually a callback now that I think of it

Marsup:
It's an API call. You can still do stuff afterwards, like tails.

PR #432, opened 2016-10-22
hapi-consolidate

Okoyl:
added hapi-consolidate under Templating category.


Marsup:
Was it not possible at all to build on top of vision like all the other template engines did ?

Okoyl:
Yes, hapi-consolidate meant to serve as an alternative for vision. I had weird issue when I tried to setup two different connections (middleware) on a single hapi server, and vision refused to work on the second connection. I could solve it by using rejoice, but I don't like the rejoice approach.

PR #436, opened 2016-11-02
Update resources.js

JKHeadley:
Hoping to be included in your list of useful resources.  Any feedback is appreciated!

### DIFF
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -71,6 +71,10 @@ exports.categories = {
             url: 'https://github.com/lynnaloo/mullet',
             description: 'Boilerplate hapi Server with Facebook and React'
         },
+        'rest-hapi': {
+            url: 'https://github.com/JKHeadley/rest-hapi',
+            description: 'A RESTful API generator built around the hapi framework and mongoose ODM.'
+        },
         rutha: {
             url: 'https://github.com/molekilla/rutha',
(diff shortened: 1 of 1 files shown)


Marsup:
This looks cool for a rapid api server, although from a quick look at the readme, hapi seems buried deep inside, I'm not sure it lets the user escape the pattern you put in place that easily.

JKHeadley:
Thanks for the feedback!  It has hapi/joi built into it's core, but the user has full flexibility to configure hapi as they need, either by adding [additional endpoints](https://github.com/JKHeadley/rest-hapi#additional-endpoints) or altering the boilerplate code.

Marsup:
Yeah I saw that, isn't it slightly late though ? As in the server already exists, I can't pass it options. How do you decide the port, whether it's http(s), ...? Basically all the decisions you have to make before getting a server.

JKHeadley:
Since this is a really new project, my initial goal was to make it easy/simple to get up and running, then slowly add in more config/customization options.  My plan is to have all/most of the configuration set up in the [``api/config``](https://github.com/JKHeadley/rest-hapi#configuration) files.  Right now that's where you can set the server port and auth type.

Marsup:
Oh ok, I read the doc too fast then :)

JKHeadley:
Np! The docs are pretty long... Thanks again for the feedback!

PR #453, opened 2017-01-14
hapi-response-time plugin

pankajpatel:
Added hapi-response-time in utility plugins list
https://github.com/pankajpatel/hapi-response-time


Marsup:
Quick feedback, `Date.now` is way faster than `getTime`, also if you want to catch very fast replies, you might be better off using [process.hrtime](https://nodejs.org/docs/latest/api/all.html#process_process_hrtime_time).

pankajpatel:
✅ Thanks for the tip; I will implement that.

PR #457, opened 2017-01-23
Remove old gh issue link, add PR info

antony:


Marsup:
@nlf do you want more user guidance on this one ? Like a small text mentioning where to put the logo, explaining the width, ...

antony:
@Marsup This is a good point - I'll add the info from the original Issue this afternoon.

nlf:
hmm, that's a good point @marsup. a little more guidance might be beneficial. maybe we should put a blurb in the readme about how to add yourself to the community page and change the link to point to that? that way we keep the section in the community page short, but we still have the details somewhere useful

thoughts?

antony:
Happy to make this change to the PR if @Marsup is also in agreement :)

AdriVanHoudt:
ping @Marsup

Marsup:
Didn't think you needed my agreement. @nlf's instructions seem to make sense.

antony:
Alright I'll jump on this when I get a second :)

antony:
Btw I'm not sure I understand the 'width' parameter myself. Going to figure out and add some extra detail.

antony:
@nlf let me know if this works for you :)

PR #459, opened 2017-02-01
Add reference to hapi-test to plugins

markkopaas:


Marsup:
The supported hapi version is over a year old, why not support more recent versions ?

PR #462, opened 2017-02-03
Update resources.js to add new boilerplate

ptpaterson:
I recently got the Angular2 quickstart example working inside of Hapi.  When I started, I struggled to find any existing information on how to do minimal Angular2 app with typescript and with Hapi.  I think that the referenced example is a good starting point and I would like to share with others.

### DIFF
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -43,6 +43,10 @@ exports.categories = {
             url: 'https://github.com/jedireza/generator-hapi-style',
             description: 'Yeoman generator for scaffolding hapi apps and plugins.'
         },
+        'hapi-angular-quickstart': {
+            url: 'https://github.com/ptpaterson/hapi-angular-quickstart',
+            description: 'Angular2 Quickstart example wrapped into a Hapi plugin.'
+        },
         'hapi-api': {
             url: 'https://github.com/rjmreis/hapi-api',
             description: 'Lean hapi API Boilerplate with an opinionated view on project structure'


Marsup:
Out of curiosity, what would make you want to do that ? https://github.com/ptpaterson/hapi-angular-quickstart/blob/master/angular-quickstart/index.js#L51
This seems useless and dangerous.

AdriVanHoudt:
Because https://github.com/ptpaterson/hapi-angular-quickstart/blob/master/angular-quickstart/index.html#L10-L13 I assume

Marsup:
If kept, this should be with very specific paths and not reveal anything running on the backend. Exposing the versions of all the modules installed on the backend is a huge security issue.

ptpaterson:
You are right of course, regarding security issues.  At one point I prefixed the name of the node_modules route with `[TOTALLY NOT SAFE]` so that it would print with the Blipp output on startup.

Knowing this, I would like to provide the most straightforward example possible, while still being a positive contribution to the Hapi ecosystem.

### Goals for example app

First and foremost, I want to help with others who are going through a process like this:

- Angular quickstart site tells me to start building an application in typescript like [this](https://angular.io/docs/ts/latest/quickstart.html).
- After reviewing Hapi vs Express vs Koa, I really want to use the Hapi framework.
- Is there anything on the internet that demonstrates how to incorporate an angular?

I am hoping that the example can be the solution for any individuals like that.  

>The are two alternative answers which are much more appropriate for production, but are much more complex, and *I* think outside of the scope of a bare bones tutorial/boilerplate app.
>1. You should use some sort of a build system that gathers all of the static content into a separate directory (e.g. /dist).
>2.  You should have two separate running servers, the Angular app, and a separate Hapi app to run a RESTful API and or other things.  Then you can follow best Angular practices directly, and do whatever you want in Hapi.

### How to deal with the current issues in the example app

I think that the code in the example should be as succinct as possible.  That means _not_ adding on an additional system (gulp, Grunt, Webpack), as well as _not_ coding a route for every node_module file (there are a dozen or so).  I think that it would be better to continue as is, but clearly mark where the issues lie.

So I would like to do the following with the example app, and PLEASE let me know if you think that this is  good and/or if more should be done.

-  Add `[SECURITY ISSUE]` as prefix to route name to clearly identify it in the Blipp output.
-  Update the README to articulate what the issues are.
-  Other..?

Thanks all!

Marsup:
I would say, if you don't want any build tool or a `cp` script of your own, the simplest way is to have something like this :
```js
const Path = require('path');
const staticRoutes = [
  'core-js/client/shim.min.js',
  'zone.js/dist/zone.js',
  'systemjs/dist/system.src.js'
].map((library) => ({
  method: 'GET',
  path: `/assets/${Path.basename(library)}`,
  handler: {
    file: require.resolve(library)
  }
}));
server.route(staticRoutes);
```
This code is untested and written in here, it's just to explain the intention, so adapt it if it fails.

Adding these routes will serve precisely the files you need (on `/assets/shim.min.js`, `/assets/zone.js` and `/assets/system.src.js`) and won't depend on how npm sets up his modules. You indeed have to add all libraries you want to use, which is cumbersome, but better safe than sorry. It also has the advantage of crashing the server straight from the start if one of your library decides to change the paths of its files, but that's just a side-effect.

Marsup:
And just to be clear, I appreciate your contribution very much, which is why I merged it. I just want to promote best practices as much as I can, and I felt this problem was important, especially so on boiler plates, since new developers might use it without understanding it fully and without knowing they would be exposed.

ptpaterson:
It is verbose, but still should be plain enough.

[https://github.com/ptpaterson/hapi-angular-quickstart/blob/master/angular-quickstart/index.js#L16-L88](https://github.com/ptpaterson/hapi-angular-quickstart/blob/master/angular-quickstart/index.js#L16-L88)

I routed it so that no changes need to be made to any of the quickstart files.  That means that some URLs are still pointing to `/node_modules/...`, but I don't want to throw managing the Angular config javascript files in there, too!

Thanks again.

PR #465, opened 2017-02-07
adding `hapi-cron` plugin reference

antonsamper:


lib/plugins.js
```diff
@@ -265,6 +265,10 @@ exports.categories = {
             url: 'https://github.com/felixheck/bissle',
             description: 'Minimalist HALicious pagination reply interface for HapiJS'
         },
+        cron: {
+          url: 'https://github.com/antonsamper/hapi-cron',
```
vdeturckheim: Can you please fix indentation here (Expected indentation of 12 space characters but found 10  indent)

lib/plugins.js
```diff
@@ -265,6 +265,10 @@ exports.categories = {
...
+          description: 'Cron jobs for internal hapi.js routes'
```
vdeturckheim: ditto

antonsamper:
@vdeturckheim good spot, all fixed

Marsup:
Just a suggestion, I think you should start the cron jobs on `onPostStart` of the server, if the other plugin fails your cron jobs will still be running, not sure it's intentional. Also you might want to stop them when the server stops.

antonsamper:
@Marsup Thanks for suggestion, ill update the plugin as it makes sense

PR #470, opened 2017-02-09
Sort npm versions with Semver.compare()

mtharrison:


lib/npm.js
```diff
@@ -30,7 +31,7 @@ exports.methods.push({
 
         return Utils.download('https://registry.npmjs.org/hapi', { json: true, rejectUnauthorized: false }, (err, body) => {
 
-            return callback(err, body && body.versions && Object.keys(body.versions).reverse());
+            return callback(err, body && body.versions && Object.keys(body.versions).sort(Semver.compare).reverse());
```
Marsup: Maybe use `rcompare` and remove the reverse then :)

mtharrison: 👍  I didn't even know that was a thing.

PR #473, opened 2017-02-27
API Docs article styling changes for easier usability.

evanperriello:
In response to #200 I made some basic changes to the article css to help with the readability of the API Documentation. I have added a left border for lis that are listed in other lis, changed the font sizing to more clearly show the structural hierarchy, dropped the bullet points to remove clutter, and added some indentation to more clearly highlight the first line of each li. This is my first open source pull request, so any feedback is more than welcome. :)

interimdocstyling


nlf:
interesting. i kinda like this, it does seem to make it a little easier to read at a glance. i'm curious to see what others have to say before i merge it though.

i almost wonder if alternating the color of the dotted lines between dark gray and light gray would make it even easier, just to make the nesting of the lists really distinguished. i'm not a ux guy though, i'm just throwing out what comes to mind. if it doesn't make sense feel free to ignore me :)

thanks for the effort @evanperriello!

AdriVanHoudt:
alternating or changing color (dark to light or something) would idd be even more helpful in deep nested parts I think

evanperriello:
I added in the alternating border colors. I feel like there's probably a less messy way to tackle this in the stylus file, but this is all that was coming to mind.

AdriVanHoudt:
I found http://stackoverflow.com/a/2477586/2339622 but not if that works and what the website should support since it is css3 :P

evanperriello:
Thanks for this, but I think that will only work in a case where they're all children of the same element, as opposed to alternating between children and grandchildren, etc.

As a side note, I hope CSS3 is okay for these doc pages--I shudder to think about anyone trying to figure out JS frameworks using IE8. :-D

nlf:
wouldn't an [nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) selector be easier for the alternating colors?

sirgallifrey:
nth-child is for siblings, in this case we have to deal with descendants, could be done like [this](http://codepen.io/sirgallifrey/pen/RpRWJp) (which is more or less the same thing in disguise) but your way is easier to understand what is happening IMO

evanperriello:
Thanks @sirgallifrey. Your stylus-fu is far stronger than mine. I'd be happy to make this change to the pull request if others feel that it's the better way to handle it.

JustinBeaudry:
I really like this addition, it makes the docs easier to follow, which is always a plus. Any reason not to merge this in?

hueniverse:
@evanperriello can you resolve the conflicts and get this ready for merging?

evanperriello:
Hey, @hueniverse, I'm pretty swamped at the moment with other projects, and for some reason I can't seem to get a local build of this working today. I can dig into it further in a couple weeks, but it seems like a recompile of the main.css file is all that's needed, so if you'd rather just close this PR, I don't mind if you just snag my few lines of code from api.styl and recompile.

evanperriello:
Okay, was able to get this to recompile, so it should be all set.

shugigo:
Hi,

I didn't notice this PR before I opened "HapiJS API Documentation Lacking Bullet Hierarchy Indication #606":
https://github.com/hapijs/hapijs.com/issues/606 

As I wrote there, in most cases you look for a phrase or a word in the entire document and it is very hard to understand the context of the match you are looking at.  
With your solution, it may ease the problem when the entire context is on the same page but not when it requires scrolling up or down.

I highly recommend adding a footer with the path that will indicate where are you.
For example: Request lifecycle -> Errors -> ...

Thanks,
Shay

Nargonath:
What is the current state of this PR? Does the change made by @evanperriello still satisfy everyone and we just need to resolve the conflicts?

hueniverse:
Let's merge it and improve later. Ping me when it is resolved and I'll merge.

evanperriello:
HI @hueniverse, This should be all set to merge. Thanks.

Nargonath:
I must admit it feels weird to have the top level bulletpoints removed when you have a long list of items, don't you think ? I do like the grey lines for the sub list though. 
image

hueniverse:
I've merged this since it is an improvement. If we want to bring back top level bullets, we can do that in a follow up change.

Nargonath:
@hueniverse fair enough.

hueniverse:
There is a history here of good incremental improvement that never get merged because they are not perfect. A lot of wasted work has gone on for last couple of years. I'd like to see that improve. Unless a PR breaks something or there are strong objections to the change, we should merge and then work to continue improving it.

Nargonath:
Alright, I will remember that.

PR #477, opened 2017-03-10
Add HTML5 document to main intro

thevangelist:


thevangelist:
I think that there are novice developers who don't know much about HTML so it's the responsibility of more experienced developers to provide correct examples.

PR #487, opened 2017-05-06
add hapi-arch to the list of plugins

AhmedAli7O1:
[npm]() [npm]()

NPM: https://www.npmjs.com/package/hapi-arch

Hapi Arch is a Hapijs plugin, with integrated CLI tool to generate a full MVC Hapi project, including one plugin and it's content e.g controllers, services, models, routes different environment configurations.

once you generate a project using ` hapi-arch generate new ` you'll be able to ` npm install && npm start ` and your server will be available to start hacking.

behind the scene, hapi arch will load all your plugins and it's controllers, services, models..etc and make them available globally " of course on the plugin scope " so you don't need to require anything, plus you got to separate your logic in an MVC way.

also you can blacklist some plugins so arch will not load them " useful in development ".

Hapi Arch, inspired by Rails, Sailsjs and Laravel.

Thanks.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -303,6 +303,10 @@ exports.categories = {
             url: 'https://github.com/nathanbuchar/hapi-glee',
             description: 'Specify environment-specific scopes for your routes'
         },
+        'hapi-arch': {
+            url: 'https://github.com/AhmedAli7O1/hapi-arch',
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
I learned that there is headers validation step (https://github.com/hapijs/hapi/blob/5b8c5ff48e8ff7b9642089a84fae42d22e182c44/lib/route.js#L246) by reading your reamde 👌

PR #491, opened 2017-06-21
Added hapi-auth-ip-whitelist to plugins list

chamini2:


Marsup:
I think you should put some work on the readme, it's a bit raw :)

PR #495, opened 2017-07-18
added, my humble plugin, hapi-acl-auth, to plugins.js

charlesread:


AdriVanHoudt:
Do you know about the `auth.access.scope` setting in the route options? https://hapijs.com/api#route-options

charlesread:
Hi, @AdriVanHoudt, I am aware of that, but in my humble opinion I think that configuration is still too tightly coupled to an authentication provider, specifically one _must_ (at least without hacking around) still use a hapi plugin to set `request.auth.creadentials.scope` while my plugin is agnostic to the very existence of `request.auth.creadentials.scope`.  Also, I didn't see where `auth.access.scope` allows one to specify hierarchies; with `auth.access.scope` one must specify all allowed roles in a route, even if it is clear that a role like "admin"  encompasses the privileges of a role like "user".

AdriVanHoudt:
👌

PR #506, opened 2017-08-30
add hapi-view-models plugin

antony:
Adds the hapi-view-models plugin we created for filtering payload data based on scopes.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -18,6 +18,10 @@ exports.categories = {
         'hapi-rbac': {
             url: 'https://github.com/franciscogouveia/hapi-rbac',
             description: 'A Rule Based Access Control module for hapi'
+        },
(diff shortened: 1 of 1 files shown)


AdriVanHoudt:
@antony I think this belongs in the Templating category no?

antony:
@AdriVanHoudt I was torn, I couldn't find the ideal category - realistically it has nothing to with templating, it's about restricting returned data based on credentials. I felt like it was closer to a security plugin than a rendering plugin, as it doesn't really provide any rendering support.

I don't know - I'm happy to move it if you feel it belongs somewhere else.

AdriVanHoudt:
@antony I must have misread then :P, in that case Authorization sounds logical to me

antony:
Not sure how I missed that - done!

PR #508, opened 2017-09-05
Update plugins.js

captainjackrana:
Added a plugin to utilities that provides easy http/https and www/non-www redirections


AdriVanHoudt:
I think it would be safer to use 
```js
Url.format({
                protocol: 'https',
                hostname: host,
                pathname: request.url.path,
            })
```
but looks great otherwise

captainjackrana:
Thanks for the suggestion.. I'll refactor the plugin and add it to the next release

PR #509, opened 2017-09-09
Added validation options to tutorials in en_us and pt_br.

joaopjt:
i was boring with the validation, because only first field with error encountered in validation was returned.

So i opened a issue in Joi repo and guys answer me with the option `abortEarly`, who make Joi return all validation erros.

With no documentation about that in hapijs, i read the validation "core", and i discovered an object assign with `options` object at `route.config.validation` object.

Here is the documentation about it :)

PS: Sorry for my english, and maybe is better to do a revision in en_us version.

### DIFF
--- a/lib/tutorials/en_US/validation.md
+++ b/lib/tutorials/en_US/validation.md
@@ -207,3 +207,32 @@ Because `response.failAction` is not specified, hapi will respond with a `500` e
 The error response will *not* indicate the reason for the error.
 If you have logging configured, you will be able to inspect your error logs for information about what caused the response validation to fail.
 If `response.failAction` were set to `log`, then hapi would respond with the original payload, and log the validation error.
+
+## Options
+
+You can use validation options to show all fields with validate errors.
+See the exemple below:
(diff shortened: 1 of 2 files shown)


JustinBeaudry:
Idk about this. I think adding a section on how to pass options is acceptable, but not this specific attribute as it's not something that every user of the library will/should be concerned with. My opinion on this is:

* add a Options section to highlight how to pass options to Joi through Hapi
* make a note about `abortEarly` and it's possible effect on validation

hueniverse:
@JustinBeaudry Our generally approach should be to accept useful additions and improve them afterwards than trying to get every submission to be exactly right. This is how we ended up with dozens of issues and PRs open for months.

I think the language should be slightly adjusted to make it more generic as you suggested but I it would probably take less time to just create an alternative PR to replace this one vs. leaving detailed notes... (e.g. you already did more of the work but "wasted" it in comments here...).

hueniverse:
Closing since no one followed up.

PR #512, opened 2017-09-29
added hapi-redis2 and hapi-mysql2

midnightcodr:
I created these two plugins because it's not easy to use the exiting `hapi-redis` and `hapi-mysql` plugins to handle situations where multiple connections are required.


midnightcodr:
Updated. Thanks @AdriVanHoudt

PR #520, opened 2017-10-25
Added hapi-graylog to plugins page

ubirajaramneto:


ubirajaramneto:
Let me know if there is any problem with this PR and I'll correct it.

PR #522, opened 2017-11-15
update contribute.jade see all GH links

rdela:
Update “See all” links on /contribute to fix 404s


rdela:
Well nevermind, links work now and I see they get all repos rather than just `hapi` so this is an unwelcome change, not sure why they were 404ing for me earlier. At least I learned how to work on the site! On to real issues. Closing PR.

PR #526, opened 2017-11-29
added plugin spazy

AlexanderElias:
added plugin spazy


Marsup:
Looks useful indeed, thanks !

PR #543, opened 2018-02-28
Update Auth Tutorial for v17

mtharrison:
To drive the updates of the tutorials, over the next couple of weeks I'm going to update all tutorials for hapi 17 unless someone else creates a PR before me. This is the first.

Closes #542

PR #544, opened 2018-02-28
Update eslint and lint config

devinivy:
I updated eslint and the eslintrc so that javascript supported by the latest versions of nodejs can be used in the codebase and documentation.  This may be applied to #543, for example, so that arrow functions in the new tutorials do not cause CI to fail.


mtharrison:
LGTM 👍

PR #546, opened 2018-03-01
Update Cookie tutorial for v17

### LINKED ISSUE #545: Update cookies tutorial for hapi 17

mtharrison:
Closes #545


marcuspoehls:
I think the part of getting a cookie value is missing. It was missing before, but we should add that.

@mtharrison What do you think?

mtharrison:
@marcuspoehls agreed, we should add it. Do you want to make a PR?

marcuspoehls:
@mtharrison yep, will submit it

PR #548, opened 2018-03-02
Update plugins tutorial for hapi 17

mtharrison:


lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
+        // ...
+    }
+};
 ```
 
-This label can then be used to apply plugins and other settings only to specific connections by using the `server.select()` method.
+Note that in the first example, we set the `name` and `version` properties explicitly, however in the second we set a `pkg` parameter with the contents of package.json as its value. Either method is acceptable.
 
-For example, to add a route only to connections with a label of `'api'`, you would use:
+Also note that when written as a module, the module should export the plugin under the `exports.plugin` property.
```
devinivy: This is useful when you export a module that isn't solely a hapi plugin, but beyond that I think this is a stylistic choice.

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
 
-```javascript
-const api = server.select('api');
+Additionally, the plugin object may contain the property `multiple` that when set to `true` tells hapi that it is safe to register your plugin more than once in the same server.
```
devinivy: I realize that you're primarily working with the existing content, but what do you think about mentioning `once`?  If I had to choose between `once` and `multiple` for a tutorial, I personally find `once` to be simpler to understand (i.e. easier to imagine a use-case) and used more often.

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
-```
+### The register method
 
-Multiple labels can be selected at the same time by passing an array of strings, this works as a logical OR statement. To accomplish a logical AND, you can chain calls to `server.select()` like so:
+As we've seen above, the `register` method accepts two parameters, `server` and `options`.
 
-```javascript
-// all servers with a label of backend OR api
-const myServers = server.select(['backend', 'api']);
+The `options` parameter is simply whatever options the user passes to your plugin. No changes are made and the object is passed directly to your `register` method.
```
devinivy: Consider mentioning `server.register()` here for context.  For example,
> ...whatever options the user passes to your plugin when registering it using `server.register()`.

devinivy: I can sort of see it both ways—passing options is covered below.  Another option is to just reference the section coming up.

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
-    }
-}, (err) => {
+const start = async function () {
 
-    if (err) {
-        throw err;
-    }
-});
+    await server.register({
+        register: require('myplugin'),
```
devinivy: Should be `plugin: require('myplugin')`

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
-}], (err) => {
-
-    if (err) {
-        throw err;
-    }
-});
+const start = async function () {
+
+    await server.register([{
+        register: require('plugin1'),
```
devinivy: Ditto here on the property name changing from `register` to `plugin`—I wont mark any others, but there are likely more.

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
+        options: {}
+    }, {
+        register: require('plugin2'),
+        options: {}
+    }]);
+};
 ```
 
 ### Plugin options
```
devinivy: I consider "plugin options" to be the name of the options passed into a plugin.  I usually refer to the options described in this section as "registration options" but I'm not 100% sure if others do the same.

lib/tutorials/en_US/plugins.md
```diff
@@ -1,205 +1,166 @@
...
-}, (err) => {
-
-    if (err) {
-        throw err;
-    }
-});
-```
-
-This allows you to attach a plugin to specific connections in a server without having to change the code of the plugin.
+Similarly the `options.routes.vhost` property will assign a default `vhost` configuration to any routes created by the plugins being loaded. More detail about the `vhost` configuration can be found in the [API reference](/api#-serverrouteroute).
```
devinivy: Is it worth also making a note about the `once` registration option?

lib/tutorials/en_US/plugins.md
mtharrison: Ok [this normalization here](https://github.com/hapijs/hapi/blob/master/lib/server.js#L367-L371) looks like it can either be top level or as `exports.plugin`.  What about rephrasing as:

> When written as a module, a plugin can either be top-level module export `module.exports = { register, options, ... }` or if you want your module to export more than a hapi plugin, it can be exported as `exports.plugin = { register, options, ... }`.

lib/tutorials/en_US/plugins.md
mtharrison: Not sure it's worth enumerating all the options in the tutorial. There's always the API docs to get more detailed reference.

mtharrison:
@devinivy think I've addressed all your comments in the latest commit.

lib/tutorials/en_US/plugins.md
devinivy: 👍

nlf:
@mtharrison you're my hero

PR #551, opened 2018-03-03
Replace hapi-sequelize by hapi-sequelizejs plugin

Nargonath:


Nargonath:
Forgot to add public data. What are the rights needed for the token to be generated as mentioned in the README?

valtlfelipe:
@Nargonath did you see there is an conflict in `lib/plugins.js` ?

Nargonath:
@valtlfelipe Indeed I may have gone a bit too fast on this one. I intended on removing danecando's plugin mention because it has not been maintained for more than a year now. Perhaps I'll have time tonight to correct the conflict and also add the `public/` data.

Nargonath:
I've updated the `lib/plugins.js` file and pushed the `public/` data but there are still conflicts on the very things I try to change so I don't see what I can do on my side. Is it due to someone having updated the plugin file before me?

Nargonath:
I just updated my PR as I noticed that I forgot to update my fork from the main the repo.

valtlfelipe:
@Nargonath there we go! Nice :)

hueniverse:
Is the other one no longer being maintained?

PR #554, opened 2018-03-04
Update sequelize plugin

valtlfelipe:
Change sequelize plugin to a updated fork with support to latest versions of hapi and sequelize


valtlfelipe:
duplicate of #551

PR #561, opened 2018-03-07
Update static files tutorial for hapi 17

### LINKED ISSUE #560: Update static files tutorial for hapi 17

mtharrison:
Changed the wording of the index/listing options too because index is actually true by default.

Closes #560

### DIFF
--- a/lib/tutorials/en_US/serving-files.md
+++ b/lib/tutorials/en_US/serving-files.md
@@ -1,6 +1,6 @@
 ## Serving static files
 
-_This tutorial is compatible with hapi v16_
+_This tutorial is compatible with hapi v17_
 
 Inevitably while building any web application, the need arises to serve a simple file from disk. There's a hapi plugin called [inert](https://github.com/hapijs/inert) that adds this functionality to hapi through the use of additional handlers.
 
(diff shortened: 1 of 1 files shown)


lib/tutorials/en_US/serving-files.md
```diff
@@ -1,91 +1,81 @@
...
+    console.log('Server running at:', server.info.uri);
+};
 
-        console.log('Server running at:', server.info.uri);
-    });
-});
+start();
 ```
 
-As you may have guessed by the option passed to the server, the `relativeTo` parameter can also be set on a per-connection or per-route level.
```
mtharrison: Maybe this is just a personal aversion, but I don't like tutorials assuming what I may have guessed by a certain point.

lib/tutorials/en_US/serving-files.md
```diff
@@ -1,91 +1,81 @@
...
+};
 
-        console.log('Server running at:', server.info.uri);
-    });
-});
+start();
 ```
 
-As I'm sure you've guessed, in its simplest form you pass a path to `reply.file()` and you're done.
+As you can see above, in its most basic form you pass a path to `reply.file()` and you're done.
```
devinivy: "pass a path to `reply.file()`" is now more like "returning `h.file()` from the handler while passing it a path"

lib/tutorials/en_US/serving-files.md
```diff
@@ -1,91 +1,81 @@
...
 
 ### Relative paths
 
-To simplify things, especially if you have multiple routes that respond with files, you can configure a base path in your server and only pass relative paths to `reply.file()`
+To simplify things, especially if you have multiple routes that respond with files, you can configure a base path in your server and only pass relative paths to `reply.file()`:
```
devinivy: I think this is the only other reference to `reply.file()` that oughtta be `h.file()` 👍

mtharrison:
Thanks, fixed those 2.

PR #575, opened 2018-03-24
Added hapi-openapi

tlivings:
Added `hapi-openapi` — a plugin for building design-driven OpenAPI apis.

I didn't add this under 'Documenting' because this plugin uses the API specification to create the routes, security, etc, as well as the documentation.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -771,6 +771,10 @@ exports.categories = {
             url: 'https://github.com/roylines/hapi-graceful-pm2',
             description: 'Handle true zero downtime reloads when issuing a pm2 gracefulReload command'
         },
+        'hapi-openapi': {
+            url: 'https://github.com/krakenjs/hapi-openapi',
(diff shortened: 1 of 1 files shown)


lib/plugins.js
```diff
@@ -771,6 +771,10 @@ exports.categories = {
             url: 'https://github.com/roylines/hapi-graceful-pm2',
             description: 'Handle true zero downtime reloads when issuing a pm2 gracefulReload command'
         },
+        'hapi-openapi': {
+            url: 'https://github.com/krakenjs/hapi-openapi',
+            description: 'Hapi plugin to build design-driven apis with OpenAPI (formerly swagger).'
```
devinivy: Only nit is that "hapi" is intended to always be styled all-lowercase.  I know there are a small handful of descriptions in here that break that, but figured it wouldn't hurt to mention!

tlivings:
Oh, yeah just figured since it is the start of the sentence. I can change it.

PR #587, opened 2018-05-31
Make directory name clear in the "Directory handler" example

bsclifton:
At first glance, it almost looks like "public" is a permission (not a directory name)


Marsup:
It's not a directory name either, it's a path, just like the name of the property mentions. Maybe `./public` would make it clearer.

bsclifton:
@Marsup good call- updated to be `directory-path-here` to make it extra clear. I think `public` is just not a good name for an example

Issue #582, opened 2018-04-21
Move "Getting started" to the top


hueniverse:
It's a bit absurd right now...

![image](https://user-images.githubusercontent.com/56631/39086355-f797a30e-4544-11e8-8036-a5ead1dcaa52.png)

ErickWendel:
I'm trying to fix, but the project isn't working. So, I remove "token" word from  internals.requestOptions.headers.authorization in GitHub.js file and now It works. 

Only `/tutorials is` with `Response Error: 401 Unauthorized` error when I access, another routes is running ok, but I don't know what only this route is not working

ErickWendel:
that issue was merged :D Can you close this issue ?

PR #596, opened 2018-06-15
move menu getting start to top

ErickWendel:
Signed-off-by: Erick Wendel <erick.workspace@gmail.com>

An observation: 

When that project calls GitHub api, with  `authorization: token `,  the API is refusing and returning 401, without `token` word, Github's api works fine. It happens on `gitHub.js` and `markdown.js` from `lib`

I think that production token is different so, I don't change api calls, only change menu.

from #582

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -108,12 +108,16 @@ module.exports = (server) => {
                     tutorial = Tutorials[language].default;
                 }
 
+                const menus = Tutorials[language].all;
+                const firstItem = 'getting-started';
+                const item = menus.filter(({ slug }) => slug === firstItem);
+                // add getting started on first position men
+                const all = item.concat(menus.filter(({ slug }) => slug !== firstItem));
                 const html = await request.server.methods.markdown.parse(tutorial.contents);
-
                 const context = {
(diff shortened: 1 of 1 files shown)


lib/routes.js
```diff
@@ -108,12 +108,16 @@ module.exports = (server) => {
                     tutorial = Tutorials[language].default;
                 }
 
+                const menus = Tutorials[language].all;
+                const firstItem = 'getting-started';
+                const item = menus.filter(({ slug }) => slug === firstItem);
```
AdriVanHoudt: When looking for 1 item in an array it is faster to user `Array.find` it does basically the same thing but stops looping when it finds the first match. ✨
In this case it won't really matter since we loop over 10 items or so, so you will never notice it :D

ErickWendel: hmmm, that's fine! I'll fix this 😄

ErickWendel:
@AdriVanHoudt I fixed, can you check ?

AdriVanHoudt:
I can (and will), I was waiting to see if anybody else had comments on this :P

PR #598, opened 2018-06-16
Remove bucker from logging suggestions - it appears to be obsolete

rob-h-w:
Hasn't been maintained since 2015, still invites users to load it using
pack.require().

I have not updated the Korean language instance of this page because my
Korean isn't up to that. Also, someone with knowledge of Portugese should
review my change.

Also remove trailing whitespace.


Marsup:
I'd rather agree but I'm going to wait for @nlf to tell if he has any intention of supporting bucker any longer.

nlf:
I don’t. I quit using it a while ago

PR #601, opened 2018-07-13
add hapi-wechat & hapi-seneca

dhso:


lib/plugins.js
```diff
@@ -246,6 +246,10 @@ exports.categories = {
         susie: {
             url: 'https://github.com/mtharrison/susie',
             description: 'Server-Sent Events for hapi with support for streaming events'
+        },
+        hapi-wechat: {
```
devinivy: This key needs single quotes (`'`) around it because it contains a dash (`-`).

lib/plugins.js
```diff
@@ -776,6 +780,10 @@ exports.categories = {
         thurston: {
             url: 'https://github.com/ruiquelhas/thurston',
             description: 'Route-level file signature validation for request payload file streams'
+        },
+        hapi-seneca: {
```
devinivy: This key needs single quotes (') around it because it contains a dash (-).

lib/plugins.js
dhso: oh, thanks.

Nargonath:
thanks @devinivy

PR #615, opened 2018-10-07
Fix #584: Api display only latest v17 and v16

### LINKED ISSUE #584: Remove unsupported versions
Basically, remove the API selections for anything other than the latest v16 and v17. We need to push people in that direction.

Nargonath:


public/css/sections/api.styl
```diff
@@ -142,7 +142,7 @@
         position: absolute
         top: 98%
         right: 0
-        height: 200px
+        // height: 200px
```
AdriVanHoudt: Why not just remove the line?

Nargonath: I thought of in case we would want to revert back at some point hence we'd still have the information of length back then that's all. I can definitely remove it.

AdriVanHoudt: i would just remove it :P

hueniverse: Drop it. Keep it clean.

Nargonath: Done

PR #626, opened 2018-11-12
addition to Hapi.js users

tnolet:


tnolet:
@hueniverse sorry for that. Updated.

PR #630, opened 2018-11-16
Update getting-started.md

dasDaniel:
Added note about host setting re. use in a docker container.

### DIFF
--- a/lib/tutorials/en_US/getting-started.md
+++ b/lib/tutorials/en_US/getting-started.md
@@ -47,7 +47,7 @@ number to listen on. After that we start the server and log that it's running.
 When creating a server, we can also provide a hostname, IP address, or even
 a Unix socket file, or Windows named pipe to bind the server to. For more details, see [the API reference](/api/#-server-options).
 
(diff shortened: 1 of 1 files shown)


lib/tutorials/en_US/getting-started.md
```diff
@@ -47,6 +47,8 @@ number to listen on. After that we start the server and log that it's running.
 When creating a server, we can also provide a hostname, IP address, or even
 a Unix socket file, or Windows named pipe to bind the server to. For more details, see [the API reference](/api/#-server-options).
 
+The host field set to `localhost` is the likely the safest choice. In a docker container, however, the localhost may not be accessible outside of the container, and using `host: '0.0.0.0'` may be needed.
```
Nargonath: Bear with me because english is not my first language but I'd write that sentence as is: The host field set to `localhost` is ~~the~~ likely the safest choice. The extra `the` feels off otherwise LGTM and I'll merge once we sort this out.

Nargonath: @dasDaniel No problem. 😉

PR #651, opened 2019-01-11
Add Softonic to hapi community listing

akira28:
I also changed permissions of 2 logos that had executions permissions (now all logos are 644)


Nargonath:
@akira28 Thanks for enriching the community! 👍

PR #656, opened 2019-02-02
Feature/testing tutorial

antony:
Add an introductory tutorial about testing with `shot`


lib/tutorials/en_US/testing.md
```diff
@@ -0,0 +1,120 @@
...
+
+process.on('unhandledRejection', (err) => {
+
+    console.log(err);
+    process.exit(1);
+});
+
+```
+
+Note that we call `server.initialize` in our `init` method. We stil want Hapi to set up all our server concerns, such as [caching](https://hapijs.com/tutorials/caching).
```
Nargonath: There is a typo: ~~stil~~ **still**

lib/tutorials/en_US/testing.md
```diff
@@ -0,0 +1,120 @@
...
+const { start, stop } = require('lib/server');
+
+start();
+```
+
+What we've created here is a way of starting the server normally by calling its `start` function in our entrypoint, and exposing a port for external HTTP traffic, but we've also got a module which doesn't do anything by default, which we can use in our tests.
+
+### Writing a route test
+
+In this example we'll use `mocha`, but the same method can be used for any testing tool.
```
Nargonath: I would rather make this example use [Lab](https://github.com/hapijs/lab) as it is our test runner from hapi's ecosystem (and [Code](https://github.com/hapijs/code) for the assertion library), but you can also keep the mention about using any other test runner (Jest, Mocha, Tap, ...).

antony: Last time I used Lab it wasn't recommended for general use so I didn't know that had changed. I've switched it now. I was already using `code` - I'm not a fan of `chai`.

lib/tutorials/en_US/testing.md
antony: Fixed

lib/tutorials/en_US/testing.md
```diff
@@ -0,0 +1,123 @@
...
+start();
+```
+
+What we've created here is a way of starting the server normally by calling its `start` function in our entrypoint, and exposing a port for external HTTP traffic, but we've also got a module which doesn't do anything by default, which we can use in our tests.
+
+### Writing a route test
+
+In this example we'll use [lab](https://github.com/hapijs/lab), but the same method can be used for any testing tool such as [Mocha](https://mochajs.org/), [Jest](https://jestjs.io/), [Tap](https://www.node-tap.org/), [Ava](https://github.com/avajs) etc.
+
+You should probably install mocha and code before trying to run this:
```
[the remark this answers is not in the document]
antony: Whoops - fixed :)

PR #658, opened 2019-02-22
Add @podium/hapi-layout and @podium/hapi-podlet plugins

trygve-lie:
[Podium](https://podium-lib.io/) is framework for writing [Micro Frontends](https://micro-frontends.org/). Podium version 3.0.0 is http framework agnostic and can now be used with multiple http frameworks, including Hapi.

This adds the plugins for writing Podium layout and fragment servers with Hapi:

 - https://github.com/podium-lib/hapi-layout
 - https://github.com/podium-lib/hapi-podlet

NB: Currently the documentation on Podium only reference Express, but will shortly also include examples for Hapi and other http frameworks.

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -863,6 +863,14 @@ exports.categories = {
             url: 'https://github.com/hapijs/nes',
             description: 'WebSocket adapter plugin for hapi routes'
         },
+        'podium-hapi-layout': {
+            url: 'https://github.com/podium-lib/hapi-layout',
+            description: 'Hapi plugin for writing Micro Frontend Layout servers with Podium - https://podium-lib.io'
+        },
+        'podium-hapi-podlet': {
+            url: 'https://github.com/podium-lib/hapi-podlet',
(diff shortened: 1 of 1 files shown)


Nargonath:
Just one thing before I merge, is it normal that both of them have the same description?

trygve-lie:
Thanks for accepting this :)

> not sure if this is the best category for it

Not sure about the category either but I did not find any of the others to really fit either. 

> Just one thing before I merge, is it normal that both of them have the same description?

Yeah kinda. One have two parts which have different roles in such a micro frontend setup. One part has the role of rendering and serving a HTML fragment. This part we call a Podlet server. 

The other part is resposible for handling input requests from end users, and requesting these HTML fragment(s) over HTTP and composing a whole page out of the different fragments and serve that as a whole page to the end user. This part we call a Layout server.

So; when building stuff with this, we always talk about building a Podlet / fragment server or a Layout server. What these do are different, but the simple description ends up being kinda the same since the difference lay in the type of server you are writing.

Nargonath:
Fair enough, thanks for the explanation.

PR #659, opened 2019-03-13
Added Hapi.js Handbook

akhromieiev:


lib/resources.js
```diff
@@ -18,6 +18,10 @@ exports.categories = {
         'hapi with Typescript': {
             url: 'http://hapibook.jjude.com',
             description: 'This book introduces Typescript into developing web applications with hapi. Typescript, a language developed by Microsoft, brings the benefits of strong types and object-oriented programming to Javascript. With ample code samples, this book teaches all aspects of developing a web application in hapi with Typescript. It also introduces, hapidock, a docker container with all required components for developing in hapi. With hapidock, you can develop and test hapi applications in a production-like environment. Through this book, you will learn to build stable enterprise-level server applications. The book is free to read at http://hapibook.jjude.com. E-books (pdf, mobi, and epub) and fully executable code samples are available after purchase.'
+        },
+       'Hapi.js Handbook': {
```
Nargonath: I just checked the error from Travis and you are missing one space character at the beginning of the line which breaks the indentation. Please add that character.

akhromieiev: Should be fixed here [https://github.com/hapijs/hapijs.com/pull/659/commits/d6ee3b695f1048e7378a894ffcba0933ff7facd8](url)

qpbp:
@Nargonath Hi! 
Is it possible to redeploy the hapijs.com site? I see an old version

Nargonath:
@qpbp The added part is not present indeed. The site deployment should be automatic. @nlf Do you notice a problem in the deployment process? Updates to the website should be fast since you fixed the  github hook last time, shouldn't they?

PR #660, opened 2019-03-18
Updated routing tutorial

jarrodyellets:


lib/tutorials/en_US/routing.md
```diff
@@ -1,61 +1,73 @@
-## Routing
+# Routing
 
-_This tutorial is compatible with hapi v17_
```
hueniverse: We need to keep this information as people come with different versions of hapi and need to know which one this applies to.

lib/tutorials/en_US/routing.md
```diff
@@ -1,61 +1,73 @@
...
+        return 'Hello World!';
     }
 });
 ```
 
-## Methods
+## <a name="methods" /> Methods
 
-The route above responds to a `GET` request to `/` with the string `Hello!`. The method option can be any valid HTTP method, or an array of methods. Let's say you want the same response when your user sends either a `PUT` or a `POST` request, you could do that with the following:
+The route above responds to a `GET` request to `/` with the string `Hello World!`. The method option can be any valid HTTP method, or an array of methods. Let's say you want the same response when your user sends either a PUT or a POST request, you could do that with the following:
```
hueniverse: Why remove the `` from the methods?

lib/tutorials/en_US/routing.md
```diff
@@ -1,61 +1,73 @@
...
 
-As you can see above, we have the string `{user}` in our path, which means we're asking for that segment of the path to be assigned to a named parameter. These parameters are stored in the object `request.params` within the handler. Since we named our parameter `user` we are able to access the value with the property `request.params.user`, after URI encoding it so as to prevent content injection attacks.
-
-### Optional parameters
+## <a name="optionalParameters" /> Optional Parameters
 
-In the above example, the user parameter is required: a request to `/hello/bob` or `/hello/susan` will work, but a request to `/hello` will not. In order to make a parameter optional, put a question mark at the end of the parameter's name. Here is the same route, but updated to make the `user` parameter optional:
+In the above example, the user parameter is required: a request to `/hello/ferris` or `/hello/cameron` will work, but a request to `/hello` will not. In order to make a parameter optional, put a question mark at the end of the parameter's name. Here is the same route, but updated to make the `user` parameter optional:
```
hueniverse: Any reason you changed all the names?

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +77,127 @@ server.route({
...
+server.route({
+    method: 'GET',
+    path: '/',
+    handler: (request, h) => {
+
+        return `Hello ${request.query.name}!`;
+    }
+});
+```
+Here, we simply access the `name` query parameter and return it in the handler, which would read `Hello ferris!`.
```
hueniverse: This should include a discussion of using the query string parser option with an example using the qs module.

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +77,127 @@ server.route({
...
+
+404 errors will happen whenever your server can't find what was a resource that was requested.  It is best practice to handle these errors the proper way.  This is easy to do in hapi, by just employing a route that will catch everything your other routes will not.  
+
+```js
+server.route({
+    method: '*',
+    path: '/{any*}',
+    handler: (request, h) => {
+
+        return h.view('404').code(404);
```
hueniverse: If using the `h.view()` method, the example should be fulling working with the vision plugin loaded.

lib/tutorials/en_US/routing.md
```diff
@@ -1,61 +1,73 @@
...
+## <a name="overview" /> Overview
+
+When defining a route in hapi, you need three basic elements: the path, the method, and a handler. These are passed to your server as an object, and can be as simple as the following:
+
+```js
 server.route({
     method: 'GET',
     path: '/',
-    handler: function (request, h) {
+    handler: (request, h) => {
```
Nargonath: Also why switching from function to arrow functions? I'm concerned that for newcomers they would use arrow functions everywhere and then loose the ability to access properties set by `server.bind`.

jarrodyellets: Makes sense. Fixing now, will resubmit in a few

lib/tutorials/en_US/routing.md
```diff
@@ -110,14 +112,57 @@ There are two query parameters here, `name=ferris` and `location=chicago`.  In h
...
+
+```js
+const QS = require('qs');
+
+server.route({
+    method: 'GET',
+    path: '/',
+    handler: function (request, h) {
+
+        const query = QS.parse(request.query);
```
hueniverse: Use the route query parser option, not manually.

lib/tutorials/en_US/routing.md
```diff
@@ -183,21 +228,50 @@ The second property is `validate`.  This allows you to set validate rules for va
...
+    server.views({
+        engines: {
+            html: require('handlebars')
+        },
+        path: '../views'
+    });
+
+    await server.start();
+    console.log('Server running on %ss', server.info.uri);
+};
```
hueniverse: Odd order of things. Why separate the server creation from the register? Also route using view should come after views manager initialized.

lib/tutorials/en_US/routing.md
Nargonath: here is the API link: https://hapijs.com/api#server.options.query.parser

jarrodyellets: Thanks...took a little digging, but was able to find it.

lib/tutorials/en_US/routing.md
Nargonath: Done

lib/tutorials/en_US/routing.md
Nargonath: added back

lib/tutorials/en_US/routing.md
Nargonath: Reverted

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +79,225 @@ server.route({
...
+Alternatively you may pass the same value to `h.response(value)` and return that from the handler. The result of this call is a response object, that can be chained with additional methods to alter the response before it is sent. For example `h.response('created').code(201)` will send a payload of created with an HTTP status code of 201. You may also set headers, content type,content length, send a redirection response, and many other things that are documented in the [API reference](/api#response-toolkit).
 
-## Config
+The `handler` option must return a value, a promise, or throw an error.
+
+Note: handlers using a fat arrow style function cannot be bound to any `bind` property. Instead, the bound context is available under [`h.context`](https://hapijs.com/api#h.context).
```
Nargonath: ~~`bind`~~ [`server.bind()`](https://hapijs.com/api#server.bind()) more explicit IMO

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +79,225 @@ server.route({
...
+                username: Joi.string().min(1).max(20),
+                password: Joi.string().min(7)
+            }
+        }
     }
 });
 ```
+The first property under `options` is `auth`.  `auth` will set the authentication configuration for the route.  Since this route is for a new user signing up, we will disable authentication.  
+
+The second property is `validate`.  This allows you to set validate rules for various request components, such as `headers`, `params`, `payload`, and `failAction`.  We use the Joi to validate the `request.payload`.  For more info, please check the [validation tutorial](/tutorials/validation).
```
Nargonath: ~~we use the Joi~~ we use [`joi`](https://www.npmjs.com/package/joi) package

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +79,225 @@ server.route({
...
+
+## <a name="missing" /> 404 Handling
```
Nargonath: That's a good call but I wonder if it is in the right tutorial. In the whole tutorial we use `hapi` as an API server as we return JSON data from our endpoints. In that section, we use it as an application server rendering templates. I'm not sure such handling belong to an API server. That's definitely not a final call, I'm open to discussion. 😃

lib/tutorials/en_US/routing.md
Nargonath: done

lib/tutorials/en_US/routing.md
Nargonath: done

lib/tutorials/en_US/routing.md
jarrodyellets: Yeah, I see your point.  I can put it in the views tutorial, or even make an error handling tutorial.

jarrodyellets: I fixed the request...left 404 there for now till we decide where it goes

lib/tutorials/en_US/routing.md
```diff
@@ -65,63 +79,226 @@ server.route({
...
+                username: Joi.string().min(1).max(20),
+                password: Joi.string().min(7)
+            }
+        }
     }
 });
 ```
+The first property under `options` is `auth`.  `auth` will set the authentication configuration for the route.  Since this route is for a new user signing up, we will disable authentication.  
+
+The second property is `validate`.  This allows you to set validate rules for various request components, such as `headers`, `params`, `payload`, and `failAction`.  We use the [joi](https://github.com/hapijs/joi) to validate the `request.payload`.  For more info, please check the [validation tutorial](/tutorials/validation).
```
Nargonath: oops didn't see this one: ~~validate rules~~ validation rules

Also either remove ~~the~~ in `we use the joi` or add *package*: `we use the joi package`

lib/tutorials/en_US/routing.md
Nargonath: @jarrodyellets do you plan on updating your PR according to this feedback?

lib/tutorials/en_US/routing.md
Nargonath: I'll ask on Slack for what other think of it and we'll see.

lib/tutorials/en_US/routing.md
jarrodyellets: This should be updated in commit 8040e4e

lib/tutorials/en_US/routing.md
Nargonath: Alright after some discussions on Slack, we decided that we'll keep the 404 handling in the routing tutorial but we'd rather not use `vision` in this tutorial. The way you wrote the 404 handling part atm is more suited for the Views tutorial so you can use it as is in that tutorial. Showcasing the catch-all routes pattern in the routes tutorial is a good addition but let's just simplify it by returning a response which is not HTML (custom string or something like that).

lib/tutorials/en_US/routing.md
Nargonath: Yes indeed for the route declaration but I think what Eran also meant is that the server creation:
```js
const server = Hapi.server({
    port: 3000,
    host: 'localhost'
});
```
should also be inside the `init` method.

jarrodyellets: Ahhh...ok, gotcha...fixing now

Nargonath: Also reading again my comment, it feels a bit harsh. Hope you didn't take it the wrong way it was definitely not intended if that's the case.

lib/tutorials/en_US/routing.md
jarrodyellets: Sounds good....will fix now

Nargonath: Cool thank you. 👍

lib/tutorials/en_US/routing.md
jarrodyellets: Lol, no worries...all good!

Nargonath: Cool. 😃

Nargonath:
Thanks for your work @jarrodyellets. 👍

PR #665, opened 2019-03-22
Update static files tutorial

jarrodyellets:


lib/tutorials/en_US/serving-files.md
```diff
@@ -171,4 +191,51 @@ server.route({
...
+    console.log('Server running at:', server.info.uri);
+};
+
+init();
+```
+The first thing we do is require both `inert` and `path`. As you will see, we will need both of these to get our file server up and running.
+
+Next, we configure `server.options.routes`. We set the location the server will look for the static files by setting the `relativeTo` option.  
+
+After our server is configured, we then register the `inert` plugin. This will allow us to have access to the `directory` handler, which will enable us to server our files. In the `directory` handler, we configure `path`, which is required, to look in the entire `public` directory which we specified in the `relativeTo` option. The second option is the `redirectToSlash` option. By setting this to `true`, we tell the server to redirect requests without trailing slashes to the same path with the missing slash.
```
Nargonath: Isn't the last sentence supposed to be:
> By setting this to `true`, we tell the server to redirect requests without trailing slashes to the same path with **those with** the ~~missing~~ trailing slash.

jarrodyellets: Def should be....don't know what i was writing, doesn't even make sense, lol.  Fixing now

PR #666, opened 2019-03-22
Update validation tutorial

jarrodyellets:


lib/tutorials/en_US/validation.md
```diff
@@ -1,14 +1,45 @@
...
+
+`npm install joi`
+
+Then, you must import it to your project:
+
+`const Joi = require('joi');`
+
+## <a name="input" /> Input Validation
+
+The first type of validation hapi can perform is input validation. This is defined in the `options` object on a route, and is able to validate headers, path parameters, query parameters, and payload data. Keep reading for examples of input validation with `joi`.
```
Nargonath: I would add a mention about the fact that you can give to `route.options.validate` either JS object or Joi object as well. The latter allows you to set options for that validation. Here is how I would phrase it:

Notes: In the below examples, you'll see that we give JS object to route `options.validate`. Be aware that the `validate` option accepts either JS object or `joi` object for its properties. The latter allows you to set `joi` options for that particular schema. Here is a partial rewrite of the [Query Parameters](#queryparams) example:
```js
options: {
    validate: {
        query: Joi.object({
            limit: Joi.number().integer().min(1).max(100).default(10)
        }).options({ stripUnknown: true })
    }
}
```
Look [here](https://github.com/hapijs/joi/blob/v14.3.1/API.md#anyoptionsoptions) for details about such options.

PR #672, opened 2019-03-27
Update server methods tutorial

jarrodyellets:


lib/tutorials/en_US/server-methods.md
```diff
@@ -1,21 +1,40 @@
...
-Server methods are a useful way of sharing functions by attaching them to your server object rather than requiring a common module everywhere it is needed. To register a server method, you call [`server.method()`](https://hapijs.com/api#server.method()). There are two different ways to call this function. You can call it with the signature `server.method(name, method, [options])`, for example:
+- [Overview](#overview)
+- [server.method()](#server.method)
+    - [Name](#name)
+    - [Method](#method)
+- [Options](#options)
+    - [Cache](#cache)
+    - [Generate a Custom Key](#key)
+    - [Bind](#key)
+- [server.methods()](#server.methods)
```
devinivy: `server.methods` is an object.  Would suggest change this to something like `server.methods` or `server.methods.myMethodName()`.
```suggestion
- [server.methods](#server.methods)
```

lib/tutorials/en_US/server-methods.md
```diff
@@ -1,21 +1,40 @@
...
 
-```javascript
+
+
+## <a name="overview" /> Overview
+
+Server methods are a useful way of sharing functions by attaching them to your server object rather than requiring a common module everywhere it is needed. To register a server method, you call [`server.method()`](https://hapijs.com/api#server.method()). There are two different ways to call this function. You can call it with the signature `server.method(name, method, [options])`, or you can call it with the signature `server.method(method)`, where `method` is an object with `name`, `method`, and `options` parameters (note that you may also pass an array of these objects).
```
devinivy: It might be nice to make mention of caching in the overview.

lib/tutorials/en_US/server-methods.md
```diff
@@ -104,9 +128,9 @@ server.method('add', add, {
 server.methods.add(5, 12);
```
devinivy: It's a little awkward because we're not in an `async` closure here, but this call method should probably `await`ed.

lib/tutorials/en_US/server-methods.md
jarrodyellets: Will do, on it

lib/tutorials/en_US/server-methods.md
jarrodyellets: I can just remove it as well...doesn't really add to the tutorial

PR #673, opened 2019-03-28
Update plugins tutorial

jarrodyellets:


jarrodyellets:
I'll work on adding some more examples and resubmit

devinivy:
Sure!  I mean, this improvement is great!  If you are going to make another pass, the example I was thinking of is this: we show how to pass plugin options when calling `server.register()`, but we never demonstrate the connection to the `options` argument of the plugin's `register()` method.  The connection is stated but never shown through an example.  I think it's useful to do this, since folks often seem to have trouble distinguishing plugin options from registration options.

jarrodyellets:
Sounds like a plan! I'll get to work on this.

PR #674, opened 2019-03-28
Update cookies tutorial

jarrodyellets:


lib/tutorials/en_US/cookies.md
```diff
@@ -1,59 +1,98 @@
...
+- [Getting a Cookie Value](#value)
+- [Clearing a Cookie](#clearing)
 
-## Configuring the server
 
-hapi has several configurable options when dealing with cookies. The defaults are probably good for most cases, but can be changed when needed.
+## <a name="overview" /> Overview
 
-To use a cookie, first you should configure it by calling [`server.state(name, [options])`](/api#-serverstatename-options) where `name` is the name of the cookie, and `options` is an object used to configure the cookie.
+When writing a web application, cookies are often used to keep state about a user between requests. With hapi, cookies are flexible, secure, and simple to use. To use cookies in hapi, you first configure the cookie by calling `server.state()`, then set the cookie via the response toolkit.
```
marcuspoehls: I’d remove the sentence starting with `To use cookies in hapi …`. The next paragraph has the same information and it seems duplicated here. I’d just remove it from the overview

lib/tutorials/en_US/cookies.md
```diff
@@ -1,59 +1,98 @@
...
     isSecure: true,
     isHttpOnly: true,
     encoding: 'base64json',
-    clearInvalid: false, // remove invalid cookies
-    strictHeader: true // don't allow violations of RFC 6265
+    clearInvalid: true,
+    strictHeader: true
 });
 ```
+In the example above, we first name the cookie `data`. We then configure the options to customize the cookie.
```
marcuspoehls: Jarrod, what do you think of adding a note that not all cookie options are required?

I like your descriptions below and the default values are typically just fine in most cases. Developers don’t have to define all options on each cookie 😃

lib/tutorials/en_US/cookies.md
```diff
@@ -1,59 +1,98 @@
...
+
+By setting `clearInvalid` to `true`, it instructs the client to remove invalid cookies.
+
+Last, we set `strictHeader` to `true` so that there are no violations of [RFC 6265](https://tools.ietf.org/html/rfc6265).
 
-This configuration will make it so the cookie named `data` has a session time-life (will be deleted by the browser when closed), is flagged both secure and HTTP only (see [RFC 6265](http://tools.ietf.org/html/rfc6265), specifically sections [4.1.2.5](http://tools.ietf.org/html/rfc6265#section-4.1.2.5) and [4.1.2.6](http://tools.ietf.org/html/rfc6265#section-4.1.2.6) for more details about these flags), and tells hapi that the value is a base64 encoded JSON string. Full documentation for the `server.state()` options can be found in [the API reference](/api#serverstatename-options).
+### <a name="options.state" /> route.options.state()
 
-In addition to this, you can further configure cookie behaviour at a route-level by specifying two properties at the route's `options.state` object:
+In addition to this, you can further configure cookie behavior at a route-level by specifying two properties at the route's `options.state` object:
```
marcuspoehls: I’d like to add a note that the route-level cookie configurations are an addition to the `server.state` config.

Is it true what I’m saying? If yes, what do you think?

lib/tutorials/en_US/cookies.md
```diff
@@ -1,59 +1,98 @@
...
+    path: '/',
+    handler: function (request, h) {
 
-In this example, hapi will reply with the string `Hello` as well as set a cookie named `data` to a base64 encoded JSON string representation of `{ firstVisit: false }`.
+        h.state('data', { firstVisit: false });
+        return h.response('Hello');
+    }
+});
+```
+In this example, we use the cookie we configured in the tutorial above. Here, hapi will reply with the string `Hello` as well as set a cookie named `data` to a base64 encoded JSON string representation of `{ firstVisit: false }`.
```
marcuspoehls: My gut feeling is that “in the tutorial above” is wrong. I’d say it’s the section above, is it?

lib/tutorials/en_US/cookies.md
```diff
@@ -73,13 +112,14 @@ The `request.state` object contains the parsed HTTP state. Each key represents t
 const value = request.state.data;
 // console.log(value) will give you { firstVisit : false }
```
marcuspoehls: This may not be true anymore when overriding the cookie value to `test`. Should we highlight the assumption that the initial cookie value wasn’t updated?

lib/tutorials/en_US/cookies.md
```diff
@@ -73,13 +112,14 @@ The `request.state` object contains the parsed HTTP state. Each key represents t
...
 ```
-
 The sample code uses the `data` cookie key from above where the related value is set to `{ firstVisit: false }`.
```
marcuspoehls: Ah, I see. Maybe we should move this additional sentence up :)

marcuspoehls:
A general question for everyone, @hueniverse @devinivy @Nargonath: do you think we should change the wording in tutorials from `we` to `you`?

I’m pro `you` in tutorials. What do you prefer?

Nargonath:
I don't mind if we change it, I'm fine with both. Maybe it would be better to do it as another set of PRs though. What do you think?

lib/tutorials/en_US/cookies.md
jarrodyellets: Good suggestion, I'll put a note in about it

lib/tutorials/en_US/cookies.md
jarrodyellets: Yeah, I can see how this could get confusing

jarrodyellets:
Yeah, I'll swap out "we" for "you"  in the PRs.  I'll resubmit the ones that have already merged as well.  Thanks for the feedback!

Nargonath:
Cool great @jarrodyellets, thanks for the great work.

PR #675, opened 2019-03-29
Update caching tutorial

jarrodyellets:


lib/tutorials/en_US/caching.md
```diff
@@ -70,13 +87,19 @@ return h.response(result).etag('xxxxxxxxx');
 
 Check the documentation of `etag` under the [response object](/api#response-object) for more details about the arguments and available options.
 
-### Server side caching
+## <a name="server-side" /> Server-side Caching
 
-hapi provides powerful, convenient server side caching via [catbox](https://www.github.com/hapijs/catbox). This tutorial section will help you understand how to use catbox.
+hapi provides powerful, convenient server side caching via [catbox](https://www.github.com/hapijs/catbox). This tutorial section will help you understand how to use `catbox`.
```
devinivy: Any reason "catbox" is displayed as `code` one place and as text in another?  I don't mind one or the other, but I think consistency is nice 👍

devinivy: Some places we've written "server-side" and other places "server side."  I'd propose we use the latter throughout the tutorial

lib/tutorials/en_US/caching.md
```diff
@@ -154,22 +179,30 @@ start();
...
+`expiresIn`states the time, in milliseconds, the cache will expire in relation to the time the item was saved in the cache. In this case, our cache will expire 10 seconds after the item was saved. 
+
+`segment` that allow you to further isolate caches within one [client](https://github.com/hapijs/catbox#client) partition. If you want to cache results from two different methods, you usually don't want mix the results together. In [mongoDB](http://www.mongodb.org/) adapters, `segment` represents a collection and in [redis](http://redis.io/) it's an additional prefix along with the `partition` option. The default value for `segment` when [server.cache()](/api#-servercacheoptions) is called inside of a plugin will be `'!pluginName'`. When creating [server methods](/tutorials/server-methods), the `segment` value will be `'#methodName'`. If you have a use case for multiple policies sharing one segment there is a [shared](/api#-servercacheoptions) option available as well.
+
+`generateFunc` is a function that will generate a new cache item if one is not found in the cache when calling `get()`. Here, our function will generate a value of two numbers added together.
```
devinivy: Not sure if we want to go into it, but there are other common situations when this will be called.  One example is that it's called "in the background" when an item in the cache exists but is found to be stale.

lib/tutorials/en_US/caching.md
```diff
@@ -154,22 +179,30 @@ start();
 ```
 If you make a request to http://localhost:8000/add/1/5, you should get the response `6` after about a second. If you hit that endpoint again the response should come immediately because it's being served from the cache. If you were to wait 10s and then call it again, you'd see that it took a while because the cached value has now been ejected from the cache.
 
-The `cache` option tells hapi which [client](https://github.com/hapijs/catbox#client) to use.
+`server.cache(options)` provisions a cache segment within the server cache facility. In this case, our policy will be using `'mongoCache'` which we created above with `server.cache`.
 
-The first parameter of the `sumCache.get()` function is an id, which may either be a string or an object with a mandatory property `id`, which is a unique cache item identifier.
+`expiresIn`states the time, in milliseconds, the cache will expire in relation to the time the item was saved in the cache. In this case, our cache will expire 10 seconds after the item was saved.
```
devinivy:
```suggestion
`expiresIn` states the time, in milliseconds, the cache will expire in relation to the time the item was saved in the cache. In this case, our cache will expire 10 seconds after the item was saved. 
```

lib/tutorials/en_US/caching.md
jarrodyellets: Yeah, that works.  Should I switch "client-side" with "client side"?

devinivy: Ah, sorry!  I meant the former.  I like "server-side" and "client-side."  I don't mind either style, though, as long as it's consistent :)

PR #680, opened 2019-04-04
Auth: replace 'we' with 'you'

jarrodyellets:


lib/tutorials/en_US/auth.md
```diff
@@ -284,24 +284,24 @@ const start = async () => {
 
 start();
 ```
-First, we need to do is register the `hapi-auth-cookie` plugin with `server.register`. Once the plugin is registered, we configure our `strategy` by calling `server.auth.strategy`. `server.auth.strategy` takes three parameters: name of the strategy, what scheme you are using, and an options object. For our strategy, we name it `session`. For the scheme, we will be using the `cookie` scheme. If you were using `hapi-auth-basic`, this parameter would be `basic`. The last parameter is an options object. This is how we can customized our auth strategy to fit our needs.
+First, you need to do is register the `hapi-auth-cookie` plugin with `server.register`. Once the plugin is registered, you configure your `strategy` by calling `server.auth.strategy`. `server.auth.strategy` takes three parameters: name of the strategy, what scheme you are using, and an options object. For your strategy, you name it `session`. For the scheme, you will be using the `cookie` scheme. If you were using `hapi-auth-basic`, this parameter would be `basic`. The last parameter is an options object. This is how you can customized your auth strategy to fit your needs.
```
Nargonath: It looks normal but I wonder why it adds it back, why was it removed beforehand?

jarrodyellets: Hmmm not sure what is going on....don't think I touched that section

Nargonath: A GitHub bug it seems

PR #681, opened 2019-04-04
Routing: replace 'we' with 'you'

jarrodyellets:


lib/tutorials/en_US/routing.md
```diff
@@ -96,7 +96,7 @@ server.route({
...
+With this configuration, a request to `/hello/john/doe` will reply with the string `Hello john doe!`. The important thing to note here is that the parameter is actually the string `"john/doe"`. That's why you did a split on that character to get the two separate parts. The number after the asterisk represents how many path segments should be assigned to the parameter. You can also omit the number entirely, and the parameter will match any number of segments available. Like the optional parameters, a wildcard parameter (for example `/{files*}`) may only appear as the last parameter in your path.
```
Nargonath: Same thing than the auth PR, is it normal that it marks it as added? Is it an unintentional re-add?

jarrodyellets: I can't figure out why its doing that...

Nargonath: Yeah I don't know might be a bug. I checked on the original page and it is definitely fine so I'll just go forward with it. Thanks

PR #714, opened 2019-05-07
fix npm stats to pull data for @hapi/hapi fix #713

### LINKED ISSUE #713: Current version on homepage is out of date
https://hapijs.com/ is currently showing - 

> Current version: 18.1.0 ● Latest update: 12 days ago ● Downloads last month: 1,017,620

even though v18.3.1 is available 
I assume it still pulls the data from npm `hapi` and not `@hapi/hapi`

yonjah:
I tried to make a quick test with running the local version and it seem to be Ok.
But since automatic testing are only linting code I can't be sure if it matches all use cases

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -6,7 +6,11 @@ exports.methods = [];
 
 exports.methods.push({
     name: 'npm.version',
-    method: () => Utils.download('https://registry.npmjs.org/hapi/latest', { json: true, rejectUnauthorized: false }),
+    method: async () => {
+
+        const result = await Utils.download('https://registry.npmjs.org/@hapi/hapi', { json: true, rejectUnauthorized: false });
+        return result && result['dist-tags'] && result.versions && result['dist-tags'].latest && result.versions[result['dist-tags'].latest];
+    },
     options: {
         cache: {
             expiresIn: Utils.fifteenMinutes,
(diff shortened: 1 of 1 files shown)


lib/npm.js
```diff
@@ -6,7 +6,11 @@ exports.methods = [];
 
 exports.methods.push({
     name: 'npm.version',
-    method: () => Utils.download('https://registry.npmjs.org/hapi/latest', { json: true, rejectUnauthorized: false }),
+    method: async () => {
+
+        const result = await Utils.download('https://registry.npmjs.org/@hapi/hapi', { json: true, rejectUnauthorized: false });
+        return result && result['dist-tags'] && result.versions && result['dist-tags'].latest && result.versions[result['dist-tags'].latest];
```
Nargonath: I wonder if it would not be better to test with an `if` statement and return a string when we don't have the information (perhaps N/D or something like that). Right now if we don't have the information it would display `false` if I'm not mistaken. What do you think?

yonjah: Not sure how it works, but I thought that what the cache is for so if the first fetch succeed following request will still hold a valid value as long as the cache hasn't expired.

But I don't have a lot of experience using hapi cache for server methods.
Looking at the methods docs and catbox code we might actually need to throw to get the catch to return a previous value. 

Do you know if that is actually the case ?

Nargonath: Hum good question. I'm not really experienced with `catbox` either to be honest. It might be affected by the options you set to `catbox` too. I think we'll stick with what you have since you mirrored what was done previously on other methods.

yonjah: regardless of cache, maybe it's better to add a fallback value in the template so if no value is returned from the method it will display N/A. 

I'll try to submit a pull request for that

PR #677, opened 2019-04-03
Changes to prevent exceeding GitHub's API rate limit

Nargonath:
- [x] Transform **Updates** page to **Changelog** page:
  - [x] Rename *Updates* mentions to *Changelog*
  - [x] Remove issues listing
- [x] Remove issues listing from **Contribute** page and replace them with links to github searches
- [x] Prevent direct calls to GitHub's API from templates
  - [x] Handle GitHub's API calls from a cron that stores data into `redis`
    - [x] Fetch `hapi` milestones
    - [x] Fetch *help wanted* issues across all repos
    - [x] Fetch *new contributor* issues across all repos.
    - [x] Fetch dynamic data for the home page
  - [x] Consume data from `redis` within templates
    - [x] Changelog template
    - [x] Home page
- [x] ~~Allow for external trigger of the data fetching for webhooks and such~~ (not anymore since new website version is under development)

Related to #676

PR #715, opened 2019-05-08
Methods cache

yonjah:
Following discussion from #714

I updated the templates to show 'N/A' if data is not available.

I looked into caching and it seem like the only way to get a cached result returned is by throwing an error.

This is a bit risky since if the cache is unavailable (first fetch) or expired the error will propagate and the server will return a 500 response.
If cache is available it will use it's value instead.

I increased the expiry value and added stale values so data will still be fresh so I think the risk of not having a valid data in cache is really small but it can happen.

I also found a minor bug where the old code specifically checked for latest v16 to display it in the dropdown for API docs.
This obviously failed when moving to `@hapi/hapi` since v16 is not published there.
I added a small workaround to also fetch the version from the old package but if we don't think the v16 API is still relevant maybe this can be removed

### DIFF
--- a/lib/npm.js
+++ b/lib/npm.js
@@ -27,14 +27,17 @@ exports.methods.push({
 
 exports.methods.push({
     name: 'npm.version',
-    method: async () => {
+    method: async (req) => {
 
-        const result = await Utils.download('https://registry.npmjs.org/@hapi/hapi', { json: true, rejectUnauthorized: false });
-        return result && result['dist-tags'] && result.versions && result['dist-tags'].latest && result.versions[result['dist-tags'].latest];
+        const result = await req.server.methods.npm.package('@hapi/hapi');
+        return result.versions[result['dist-tags'].latest];
     },
     options: {
         cache: {
-            expiresIn: Utils.fifteenMinutes,
+            dropOnError: false,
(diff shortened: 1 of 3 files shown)


Nargonath:
Thanks for your PR. 

Regarding the update in the way we cache stuff I'm not sure we should go with it right now, even though that's definitely a good idea. There is an ongoing PR (#677) that I'm working on to refactor the way we handle most of our cached data. Instead of caching directly from the methods themselves, we want to setup a CRON that would fetch the data on a regular basis and store the values in `redis`. The server methods would then directly consume the data from redis and not perform requests themselves. The main point of this is to fix the issues we got every now and then about GitHub API rate limit exceeded but I guess it would also address the issue you are pointing out.

Concerning the `v16`, we still want to display the API documentation for now. Would you mind opening another PR for the `v16` fix or trim from this one the part about the cache system, please?

yonjah:
I've created a new branch for only those changes -
https://github.com/hapijs/hapijs.com/pull/716

Nargonath:
Cool thank you. 👍

Nargonath:
I'll just close the PR since it seems that you have extracted everything from the cache related code. Feel free to reopen if that's not the case.

PR #717, opened 2019-05-10
improve: npm-stat link to show stats for both @hapi/hapi and hapi

yonjah:
Link on home page goes to http://npm-stat.com/charts.html?package=hapi which shows download stats for old package.
Change to show both package stats.
We'll probably want to switch to only show @hapi/hapi once it picks up.


Nargonath:
Thanks, good catch.

PR #719, opened 2019-05-14
Removed extra s from the log line

liam056:


Nargonath:
Thanks for the fix. 👍

PR #720, opened 2019-05-20
add --save to install command

kwaping:


nlf:
npm saves by default and has for quite a while now, this flag isn't necessary

kwaping:
Thanks @nlf, I wasn't aware of that!

PR #725, opened 2019-07-15
fixing cookie auth example code

grawk:
In going through the auth examples, I came across a couple issues with the cookie code.

1. add `auth:false` to the login POST route. Otherwise kept getting a 302 back to login GET
2. access `id` and `password` from account object instead of non-existing-in-scope user object


Nargonath:
Thanks for fixing that. 👍

PR #727, opened 2019-07-25
Broken Link fixed

rummanhasnayeen:
Fixed link: 
- faketoe link fixed.

Not available links:
- kilt
- @watchup/hapi-mongoose
- hapi-level-db

### DIFF
--- a/lib/plugins.js
+++ b/lib/plugins.js
@@ -477,10 +477,6 @@ exports.categories = {
             url: 'https://github.com/sibartlett/hapi-io',
             description: 'A socket.io plugin that can forward socket.io events to hapi routes'
         },
-        'hapi-level-db': {
-            url: 'https://github.com/maxnachlinger/hapi-level-db',
-            description: 'HapiJS / LevelDB integration'
-        },
         'hapi-magic-filter': {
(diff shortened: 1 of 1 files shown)


lib/plugins.js
```diff
@@ -830,7 +822,7 @@ exports.categories = {
             description: 'A configuration document format, an API, and a foundation for A/B testing'
         },
         faketoe: {
-            url: 'https://github.com/hapijs/faketoe',
+            url: 'https://github.com/drummel/faketoe',
```
Nargonath: Not sure that's a good idea to add a link to that repo. `faketoe` package on `npm` is still owned by Eran etc. : https://www.npmjs.com/package/faketoe. Plus the person who forked the repo didn't add a link to another package and doesn't seem either to maintain it so I would just go and delete it from the list like the rest you deleted.

rummanhasnayeen:
I have deleted faketoe like the others. Please review changes.

lib/plugins.js
rummanhasnayeen: i have deleted faketoe like the rest. Please review changes

PR #731, opened 2019-07-26
Update caching.md

Death-Gear:
Grammatical mistake fixed
don't want mix the results >> don't want to mix the results


Nargonath (approved):
Thanks. 👍

